PR #57, opened 2018-09-08
IsBlank, In and StartsWith delegable support updated

scottdurow:
I've run some tests and the CDS and Dynamics 365 CE connectors do support delegation of IsBlank, StartsWith and In from inside a Filter query.
E.g. 
```Filter(Accounts, StartsWith(name,TextSearchBox1.Text), IsBlank(address2_city))```
will result in the fetchxml
```
<filter type="and">
		<condition attribute="name" operator="like" value="contoso%" />
		<filter type="or">
			<condition attribute="address2_city" operator="null" />
			<condition attribute="address2_city" operator="eq" value="" />
		</filter>
</filter>
```

and also:
```Filter(Accounts, AccountSearchText.Text in name)```
with result in the fetchxml
```
<filter type="and">
	<condition attribute="name" operator="like" value="%contoso%" />
</filter>
```


AFTOwen:
Thanks, @scottdurow, for the feedback!
@lancedMicrosoft Does this match your understanding? Do CDS for Apps and Dynamics support delegation of these functions?

lancedMicrosoft:
The detailed CDS delegation table should go out today.

AFTOwen:
@scottdurow Closing this PR, now that we've removed the topic that it updates and moved the information that's still relevant to the [delegation overview](https://docs.microsoft.com/powerapps/maker/canvas-apps/delegation-overview#delegable-data-sources).

PR #59, opened 2018-09-12
Updating FAQ for gateway H/A information

BrianTJackett:
Updating answer to question about enabling gateway high availability to reflect using clustering as of Nov 2017 release to enable this.  If there is a better link for pointing users on how to configure that would be better than the blog post I pointed out.


AFTOwen:
@BrianTJackett Thanks for the suggestion!
@linhtranms OK to merge this PR?

AFTOwen:
These changes are live. Thanks, @BrianTJackett!

cc: @linhtranms

PR #72, opened 2018-10-04
Update cognitive-services-api.md

VikramPendse:
Change "I" with "we" to maintain the context/flow of the document as third party reader.


AFTOwen:
Thanks, @VikramPendse!

PR #94, opened 2018-11-03
Modify "PowerApps Admin center" and URL

Dynamics365CE:
Change "PowerApps Admin center" to "PowerApps", and correct URL, since Data/Connections doesn't appear in the PowerApps Admin center.


jimholtz:
Content updated and published to live via my working branch. Thanks for letting us know!

PR #116, opened 2018-11-16
Hide trash icon when change form is in New mode

johnelsbree:
Without this change, the program has a serious data-loss bug: If a user clicks the trash icon while the form is in New mode, the app deletes an existing record (whichever record was most recently selected) from the table!

To prevent this, we should hide the trash icon when using the ChangeScreen to add a new record.


johnelsbree:
Alternatively, we could set the DisplayMode of the trash icon to **If(EditForm1.Mode = FormMode.Edit, DisplayMode.Edit, DisplayMode.Disabled)** and also adjust its DisabledColor.

Not sure which design approach would be preferred.

PR #123, opened 2018-11-17
typo: assoicated -> associated

nschonni:


KumarVivek:
The typo is in the Message definition so we just cannot update the docs:

https://docs.microsoft.com/en-us/dotnet/api/microsoft.xrm.tooling.packagedeployment.crmpackageextentionbase.importextension.isroleassoicatedwithteam?view=dynamics-deployer-ce-9

I will ask the feature team to fix the method name. Closing the PR without merging.

PR #149, opened 2018-12-19
Update use-powerapps-checker.md

RyanWAnderson:
Addressed and removed a comment in a table because it was breaking rendering.


Mattp123:
Fixed via https://github.com/MicrosoftDocs/powerapps-docs-pr/pull/1089.

PR #179, opened 2019-02-01
Update function-set.md

lancedMicrosoft:
See https://msazure.visualstudio.com/OneAgile/_workitems/edit/1471843  and https://powerusers.microsoft.com/t5/General-Discussion/Question-about-the-Set-function/m-p/51047#M21281


gregli-msft:
I hear the comments and understand how that paragraph can be confusing.  But I'd still like to have some statement there about the pitfalls of using state variables.  How about this: 

PowerApps are based on formulas that automatically recalculate as the user interacts with an app.  Any formulas that depend on a variable will automatically update when it changes.  However, the variable will not be automatically updated if the value of the formula used in the **Set** function changes.  This requires the author to manually update the variable which can be error prone and harder for others to understand.  Before you use a variable, review [working with variables](../working-with-variables.md).

lancedMicrosoft:
That works.

PR #183, opened 2019-02-03
Number of characters for Mid function is optional

carlosfigueira:
Update Left/Right/Mid documentation to state that the number of characters parameter is not required.


AFTOwen:
cc: @gregli-msft

PR #322, opened 2019-04-12
Adding key properties that pertain to Icon only - Icon, Color, Rotation.

SpencerLynn:
Adding key properties that pertain to Icon only - Icon, Color, Rotation - in a new section "Icon only key properties".


SpencerLynn:
@emcoope-msft

PR #412, opened 2019-05-24
Registration link should be updated to Partner Center

qianw211:
The link points to the nomination form. It should point instead to https://partner.microsoft.com/en-us/dashboard/account/v3/enrollment/introduction/azureisv

Please approve!


KumarVivek:
@qianw211 May I know why are you proposing this change? I cannot take this change as we want to link to our form and not a generic link.

qianw211:
@KumarVivek The old registration form is no longer used, instead partners are recommended to open an account with the Partner Center, and those still using the cloud partner portal are then being directed there.  We need to update the link to reduce live support for partners trying to publish offers in the Marketplace.

PR #417, opened 2019-05-27
chore: Generate Redirect file from MS Docs Pack

nschonni:
One article had the redirect metadata to itself and was removed


KumarVivek:
Sorry, we can't take these changes to config json files as we are changing a lot of things internally. Closing it without merging.

PR #433, opened 2019-06-05
Update use-solution-explorer.md

davidyack:
Suggest clarifying that both the CDS and the CDS (Current Environment ) connectors work when started from a solution, but neither work if the flow is outside the solution


Mattp123:
@davidyack - Sorry for the delay on getting back. The earlier version of Common Data Service is no longer available and has been replaced now for some time by the current Common Data Service and Common Data Model. Closing PR without merge.

PR #460, opened 2019-06-22
fix: cc-feature-included-with-update-8-1-0-online-only

nschonni:
Add missing file extension. File seems unreferenced either way


KumarVivek:
Closing without merging; shouldn't rename includes files

nschonni:
@KumarVivek there is no extension on the file, so it wouldn't work as an include (if it was referenced anywhere)

PR #511, opened 2019-07-11
entityReference.name is not entity type

kenakamu:
entityReference.name is not entity type


KumarVivek:
@Nkrb Can you review this.

Nkrb:
@kenakamu Thanks for the feedback. We will update the docs with the latest changes in the next release.

PR #512, opened 2019-07-11
onButtonClick sample is older than download

kenakamu:
onButtonClick sample is older than download


KumarVivek:
@Nkrb Can you review this..

Nkrb:
@kenakamu Thanks for the feedback. We have updated the docs and pushed to live. Closing this PR.

PR #514, opened 2019-07-11
code is older than sample

kenakamu:
and missing import {IInputs, IOutputs} from "./generated/ManifestTypes";


Nkrb:
@kenakamu Thank you for  the feedback. We have updated the information in other PR and pushed to live. Closing the PR.

PR #515, opened 2019-07-11
sample is older than download

kenakamu:
sample is older than download


kenakamu:
what should i do for this one? Could you give me a bit more clear instruction how to proceed please :)?

Nkrb:
@kenakamu Thanks for the feedback. We have updated the information in another PR and pushed to live. Closing the PR.

PR #605, opened 2019-08-28
 Upper case API as per interface

Davesmall28:
Change context.webApi to context.webAPI to bring inline with the actual implementation of the interface.


Nkrb:
@Davesmall28 Thanks for the feedback. We have updated the information and pushed to live. Closing this PR.

PR #678, opened 2019-10-01
Add global optionset query example

plohm12:
Add example HTTP request demonstrating how to query a single global optionset without knowing its metadata ID.


KumarVivek:
@JimDaly Can you please review/approve this change.

JimDaly:
Adding this example directly, so closing this.

PR #682, opened 2019-10-02
Changed setting reference

rappen:
*Data row limit for non-delegable queries* is now found under *Advanced settings*, not experimental anymore.


KumarVivek:
@rappen Thanks for the update.

PR #686, opened 2019-10-03
Update expense-report-install.md

yijw2017:


KumarVivek:
@yijw2017 Thanks for the correction.

PR #721, opened 2019-10-17
Fixes #720

### LINKED ISSUE #720: "Use the Solution Configuration page" add Note that it only works in classic mode
Under the heading "Use the Solution Configuration page" it says: _"This will cause a new Configuration node to appear in the Solution window underneath the Information node and above the Components node."_

This is true for the classic experience, but not for the new/modern experience. The solution configuration page currently can't be accessed from the new experience. A note should be added that says so.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: a9e73eed-f196-83d6-96d7-9d7244a945f3
* Version Independent ID: 3307cefd-8ce7-8836-4705-ed0d6dad06c2
* Content: [Create, export, or import an unmanaged solution (Common Data Service) - PowerApps](https://docs.microsoft.com/en-us/powerapps/developer/common-data-service/create-export-import-unmanaged-solution)
* Content Source: [powerapps-docs/developer/common-data-service/create-export-import-unmanaged-solution.md](https://github.com/MicrosoftDocs/powerapps-docs/blob/live/powerapps-docs/developer/common-data-service/create-export-import-unmanaged-solution.md)
* Service: **powerapps**
* GitHub Login: @shmcarth
* Microsoft Alias: **jdaly**

DennisGaida:
Add note that classic mode is necessary


phecke:
@DennisGaida Thanks for catching that :)

PR #725, opened 2019-10-18
don't confuse using directives and statements

mrlacey:
Address incorrect use of ["using statement"](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/using-statement) to refer to a ["using directive"](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/using-directive).


KumarVivek:
@mrlacey Thanks for submitting the corrections; accepting the changes.

Issue #376, opened 2019-05-08
401 Unauthorized with secret


Ballzer0:
when im using clientID and secret to generate an Authentication token, i get a 401 Unauthorized response back from the CDS api when i try to send a query

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 62419bb8-985c-cb11-ac8d-2d018868c2a6
* Version Independent ID: db669ed7-d87b-c359-f5fb-3f2f1052281d
* Content: [Use OAuth with Common Data Service (Common Data Service) - PowerApps](https://docs.microsoft.com/en-us/powerapps/developer/common-data-service/authenticate-oauth)
* Content Source: [powerapps-docs/developer/common-data-service/authenticate-oauth.md](https://github.com/MicrosoftDocs/powerapps-docs/blob/live/powerapps-docs/developer/common-data-service/authenticate-oauth.md)
* Service: **powerapps**
* GitHub Login: @paulliew
* Microsoft Alias: **jdaly**

paulliew:
are you using Application User as your client?  If so, make sure that the application is in your AAD.  And also make sure that there is a security role assigned to the Application user.

KumarVivek:
@phecke Can you please look into this issue and reach out to Paul if required.

Ballzer0:
I have registered a user in my AAD, but not given it any licenses, i have registered the same user in dynamics as an "application user" and given it permissions to read pretty much every field. 

I think it might be a problem with the domain im using for the application username as described here.
https://github.com/MicrosoftDocs/powerapps-docs/issues/377#issue-442083357

paulliew:
You shouldn't be able to registered a regular AAD user as an Application User because Application User requires an AAD application ID.  How did you register the regular AAD user as Application user?

Ballzer0:
well thats what i was unsure about, according to this documentation you should first register a user in AAD, then you should create an application user. 
but i get caught up on step 4 "Make sure the username contains the organization domain URL of your D365 tenant"
![image](https://user-images.githubusercontent.com/28560729/57507102-0dc84880-72fe-11e9-970f-618e08f8c9a5.png)


https://docs.microsoft.com/en-us/powerapps/developer/common-data-service/use-single-tenant-server-server-authentication#bkmk_registration

Ballzer0:
[...]

PR #760, opened 2019-11-01
Removed common endpoint from application secret auth code

ZachEdwards7:
The common endpoint does not support authentication with only clientId and secret. Because the application can exist in multiple tenants, there is no "home tenant" for the common endpoint to authenticate against. We have repro'd this in our tenant, and the fix was to remove /common/ and use the full Tenant ID, or the "tenant.onmicrosoft.com" equivalent

See also: https://github.com/MicrosoftDocs/powerapps-docs/issues/376


KumarVivek:
@phecke Can you pls review this change.

phecke:
@ZachEdwards7 Thanks for finding that doc bug.

PR #761, opened 2019-11-01
Editing for task 1623158

ReneeW-CPub:


ReneeW-CPub:
In particular, see my changes to the list of allowable file name characters. I made it user friendly, but for example if Azure says "digits" instead of "numbers" it would be better to use what Azure users are used to.

PR #841, opened 2019-12-05
Link fix at powerapps-docs/developer/common-data-service/overview.md

sadikkuzu:
Developer Guide for Dynamics 365 Customer Engagement (on-premises)


sadikkuzu:
https://docs.microsoft.com/en-us/powerapps/developer/common-data-service/overview

PR #862, opened 2019-12-12
Remove instructions to create and use user from AAD

RedsGT:
Microsoft keeps getting support cases where customers follow this documentation, use a real user from AAD, then as some point in the future the data from the real user syncs down from AAD and overwrites the app registration information in the user record and breaks authentication.  This change fixes the documentation and will deflect those cases.


KumarVivek:
@RedsGT Thanks for submitting a change request. Are you a Microsoft employee? If yes, we can follow up internally on this change?

@phecke Can you please review the changes, and accept/update as required.

PR #880, opened 2019-12-16
Updated to V2 of Global

davidyack:
This is the only way to get new fields like EnvironmentId


KumarVivek:
@davidyack Thanks for the updates. Accepting the change.
@phecke Can you please ensure we replicate this change throughout.

PR #941, opened 2020-01-12
No longer limited to one per form

davidyack:
You can now enable more than one field to have a canvas app


Mattp123:
@davidyack - sorry for the delay. You are indeed correct and this has been handled via https://github.com/MicrosoftDocs/powerapps-docs/pull/1668. Closing without merging.

PR #1037, opened 2020-02-16
Can we delete this article?

satishreddyv:
Not relevant in Unified Interface apps. Proposing to delete it.


KumarVivek:
@satishreddyv We still have users on the legacy web client so we can't delete this topic yet.
Closing the PR without merging.

PR #1038, opened 2020-02-16
Can we delete this article too?

satishreddyv:
Not relevant in Unified Interface apps. Proposing to delete it.


KumarVivek:
@satishreddyv We also have users on the legacy web client so we cannot delete the topic. Closing the issue.

PR #1046, opened 2020-02-18
Fix typo

wgeorgecook:
"informatiuon" should be "information" under dateFormattingInfo description.


wgeorgecook:
Small change to correct a typo. Thanks for all the hard work y'all.

PR #1172, opened 2020-03-23
correcting typo for OnSelect

HeidiHasting:
correcting typo from 'OnSeelct' to 'OnSelect'


tapanm-MSFT:
Thank you for the feedback. The doc is fixed with a few additional updates in a separate PR.

PR #1238, opened 2020-04-08
Fix some issues in the DateValue function docs

carlosfigueira:
Found a few inconsistencies between the docs and the reality, this should address them


gregli-msft:
Moved these changes to the private repro: https://github.com/MicrosoftDocs/powerapps-docs-pr/pull/2650

PR #1293, opened 2020-04-22
Add alternative Postman Environment configuration

gregdegruy:


JimDaly:
@gregdegruy 
Thanks for submitting this contribution. Sorry for the delay in responding.
I'm afraid I'm not going to accept it though.

The scope of this topic is very limited. We are not trying to document everything you can do with Postman.

This topic is about how people can set up a postman environment using the specific globally approved client id we provide: (51f81489-12ee-4a9e-aaae-a2591f45987d) and the configured callback URL so that they can being to use Postman immediately without the effort of registering an app.

The information you are providing is valuable if the have their own registered app that they want to use. But why even do that when we are providing one that works?

The doc already calls out the possibility to use their own:

> This environment uses a client ID for an application that is registered for all Common Data Service environments.
> 
> You can use the clientid and callback values supplied in these instructions. However, when building your own application, you should register your own Azure Active Directory (Azure AD) application.
> 
> To register your own Azure AD application, see the steps described in Walkthrough: Register a Common Data Service app with Azure Active Directory.

I don't really want to go any further than this.

Thanks, I hope you understand

Jim

gregdegruy:
Hello Jim, thank you for clarifying and for the detailed response.

PR #1316, opened 2020-04-28
Fixed typo in link

rappen:
The aka.ms url for section *Configure service connection for Power Apps checker* points to an invalid anchor since the space was introduced between Power and Apps.


phecke:
@rappen Thanks for the feedback

Issue #1320, opened 2020-04-29
Unified Interface transition date confusion


imranakram:
In another post the transition date is said to be December 1, 2020 as in:
"We're locked to move away from the legacy web client by December 1, 2020"

https://docs.microsoft.com/en-us/powerapps/maker/model-driven-apps/faqs-transition-unified-interface

But here it mentions October 1, 2020 still, could you please update this document and put the correct date (whichever is applicable)

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: fa292e63-df60-bab7-b858-9a71ee90a69e
* Version Independent ID: 2a6e3665-9d4d-19f4-6915-bb34eb983942
* Content: [Checklist: Unified Interface transition - Power Apps](https://docs.microsoft.com/en-us/powerapps/maker/model-driven-apps/checklist-transition-unified-interface#feedback)
* Content Source: [powerapps-docs/maker/model-driven-apps/checklist-transition-unified-interface.md](https://github.com/MicrosoftDocs/powerapps-docs/blob/live/powerapps-docs/maker/model-driven-apps/checklist-transition-unified-interface.md)
* Service: **powerapps**
* GitHub Login: @Mattp123
* Microsoft Alias: **haybass**

KumarVivek:
@imranakram This is fixed now. Closing the issue.

PR #1321, opened 2020-04-29
#1320: Corrected transition date to Unified Interface

imranakram:
Small but important change :)


imranakram:
Added basic example for lookup control as well. This is an additional commit

Nkrb:
@imranakram Thanks for the example Imran. We have added the example and is live. Closing this PR.

PR #1420, opened 2020-05-20
Explain common error

MarkMpn:
Instructions for enabling the endpoint are already linked to, but adding in the error message that's encountered when a user misses this step to make these instructions more easily findable.


KumarVivek:
@MarkMpn As you mentioned, we have already mentioned how to enable this feature so adding this note again seems redundant. So, I am closing this PR without merging.

Thanks for contributing to our docs/

PR #1425, opened 2020-05-20
Demonstrate alternative to missing partylist attributes

MarkMpn:
Instead of just listing that `partylist` attributes are not supported, illustrate how to access the same data with a different format of query.


phecke:
@MarkMpn Thanks for the suggestion, Mark.

PR #1426, opened 2020-05-20
Indicate that plugins are not fired by SQL queries

### LINKED ISSUE #1403: SQL queries return different result sets to FetchXML
It seems that plugins and rules are not being applied when I query data via TSQL rather than using a FetchXML report. As a test I have a preoperation plugin on multiple retrieve that is guaranteed to return zero results.

When that plugin is enabled a FetchXML request returns zero results, the TSQL request returns all records.

I can provide plenty of use cases where consistency is essential here - happy to provide them if needed.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 60540f3d-b959-2daa-23a1-2938de01919d
* Version Independent ID: 07b12c3a-a0db-47f1-1aca-04008c9efca9
* Content: [Use SQL to query data (Common Data Service) - Power Apps](https://docs.microsoft.com/en-us/powerapps/developer/common-data-service/cds-sql-query)
* Content Source: [powerapps-docs/developer/common-data-service/cds-sql-query.md](https://github.com/MicrosoftDocs/powerapps-docs/blob/live/powerapps-docs/developer/common-data-service/cds-sql-query.md)
* Service: **powerapps**
* GitHub Login: @mayadumesh
* Microsoft Alias: **pehecke**

MarkMpn:
Resolves #1403


phecke:
@MarkMpn I understand what you are doing here, but we cannot make that statement. The feature is in preview and as such not yet complete. Additional functionality may be added at any time.

JimDaly:
I think it is correct to add this.
It is important for people to understand this and it will not suddenly appear.

phecke:
@MarkMpn Thanks for your suggestion, merged.

PR #1452, opened 2020-05-26
Update restore-an-app.md

russgove:


tapanm-MSFT:
Thank you for reaching out. While it’s possible today using PowerShell, this may change. Hence, please consider documented 6 months limit when working with restoring apps.

PR #1473, opened 2020-06-03
Fix breaking error in sample code

fordosa90:
- Fix broken html tag blocking rendering of the code (closing tag of <td> instead of <th>)
 - Additionally included wrapper div for pagination based on OOB entitylist render, though that is optional change

### DIFF
--- a/powerapps-docs/maker/portals/liquid/render-entity-list-current-page.md
+++ b/powerapps-docs/maker/portals/liquid/render-entity-list-current-page.md
@@ -140,7 +140,7 @@ Render the Entity List associated with the current page as a paginated sortable 
                 {% else %}
                   {{ attr }}
                 {% endif -%}
-              </th>
+              </td>
             {% endfor -%}
 
             <td>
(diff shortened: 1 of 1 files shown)


tapanm-MSFT:
Thank you for your feedback. Will update the documentation accordingly in a separate PR.

fordosa90:
Thank you @tapanm-MSFT 
Out of curiosity - checking all other closed PRs in this repository, seeing them being merged - why will be the documentation be updated in this case with a different PR? 
If it is not expected to submit a PR, it could have saved time for me instead of providing a fix, simply raising a ticket.

tapanm-MSFT:
Thanks for your query. This was also part of another content issue. You can also raise GitHub issues for documentation. When a PR is raised, it needs to be reviewed and if an additional change is needed, or if it becomes a part of additional content updates, it can be updated separately. And you're correct about raising support service requests for issues with the feature discussed in the article. Hope this clarifies.

PR #1490, opened 2020-06-09
Clarify x-ms-dynamics-organization header

MarkMpn:
The `x-ms-dynamics-organization` header contains the full domain name of the instance that the webhook is triggered from, not the name of the associated tenant


JimDaly:
@MarkMpn 
Thanks for pointing this out. You are correct.
Sorry it has taken so long, but I have updated the topic directly with another change, so these changes aren't needed now.
The updated topic should appear in the public docs this evening.

PR #1495, opened 2020-06-10
Fix the Content-Range header

Kopiczek:
When providing the Content-Range header during upload, it actually has to be prefixed with "bytes ", otherwise CDS will return code 416.


phecke:
@Kopiczek Thanks for catching that error.

Issue #1509, opened 2020-06-15
Expand on JObject reasoning


MarkMpn:
This page includes the "Important" note:

> Due to certain Service Bus optimizations, it is not recommended that .NET developers deserialize the JSON formatted message request body to a RemoteExecutionContext object. Rather, use JObject to parse the message body.

However, it's not clear what the issue is that this might cause or whether/how it applies when using a webhook instead of Azure Service Bus. Can you please expand on this point so I can understand the reasoning behind it?

(this note was added in f58314291426b5f03b1207f1d0ca7047395268b3)

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: d4dfc2a3-e9b6-7c53-5bc9-a7e9eda29618
* Version Independent ID: 1f4b68f8-704a-7c6a-fc2b-2cee577d8e70
* Content: [Use webhooks to create external handlers for server events (Common Data Service) - Power Apps](https://docs.microsoft.com/en-us/powerapps/developer/common-data-service/use-webhooks)
* Content Source: [powerapps-docs/developer/common-data-service/use-webhooks.md](https://github.com/MicrosoftDocs/powerapps-docs/blob/live/powerapps-docs/developer/common-data-service/use-webhooks.md)
* Service: **powerapps**
* GitHub Login: @JimDaly
* Microsoft Alias: **jdaly**

JimDaly:
Thanks for your feedback. 
Whenever we document something we think about what information people need and try to trim it down to what is relevant.

In this case, the guidance is to help people be more successful. We noted people struggling with this, so this is our recommendation.

The reason why may be interesting to some, but since there isn't anything we expect a developer can do to change things, we haven't provided that information.

Probably not the answer you are looking for.

MarkMpn:
OK, can I ask a more specific question then:

This note mentions there is some problem deserializing the JSON within a service bus. Does this also apply to an HTTP webhook that's not using Azure Service Bus?

JimDaly:
I am not aware of any issues in that context.

The doc was updated more than a year ago, so it is very time consuming to find additional details if they weren't included in the note when it was added. But since it was very specific to the Service Bus scenario, I have to conclude that the same problem isn't known to apply to the webhook case.

PR #1521, opened 2020-06-19
Removed note specific to Azure Service Bus

MarkMpn:
As per #1509, the note about deserializing the webhook payload to a `RemoteExecutionContext` was specific to Azure Service Bus and not HTTP webhooks.


phecke:
@MarkMpn Thanks for catching this error.

PR #1553, opened 2020-06-29
Fixed plug-in time out duration in milliseconds

yasirsafeer:
Added a missing '0'.


phecke:
@yasirsafeer Thanks for catching that missing zero.

PR #1563, opened 2020-07-01
Fixed the description of the And function

github129:
The And function is also available with exactly no arguments, in which case In that case, the And function returns true, so the previous description is incorrect.


tapanm-MSFT:
Thank you for reaching out. "*all of its arguments are **true***" is applicable since the boolean logic function needs arguments for comparison and to return a value. Without arguments, there's no logic to check. For issues with the product or function behaviors, please work with support teams using a [support ticket](https://admin.powerplatform.microsoft.com/support).

PR #1592, opened 2020-07-15
Clarify use of column comparison

MarkMpn:
Provides examples for using `ConditionExpression` to compare columns and clarifies the language around its use


NHelgren:
Approved

phecke:
@MarkMpn Thanks for the review and changes :)

PR #1598, opened 2020-07-17
Update addCustomView.md

ericregnier:
Corrected the notes about invalid GUIDs

### DIFF
--- a/powerapps-docs/developer/model-driven-apps/clientapi/reference/controls/addCustomView.md
+++ b/powerapps-docs/developer/model-driven-apps/clientapi/reference/controls/addCustomView.md
@@ -32,7 +32,7 @@ Lookup
 
 - **viewId**: String. The string representation of a GUID for a view.
     > [!NOTE]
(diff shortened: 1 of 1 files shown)


JimDaly:
Since the change directly changes the meaning of the current description. Need feature team to approve.

JimDaly:
@miglisic & @chmoncay 

Change to [addCustomView (Client API reference)](https://docs.microsoft.com/en-us/powerapps/developer/model-driven-apps/clientapi/reference/controls/addcustomview) directly contradicts current documentation saying it doesn't work.

**Current:**
> This value is never saved and only needs to be unique among the other available views for the lookup. A string for a non-valid GUID will work, for example “00000000-0000-0000-0000-000000000001”. It’s recommended that you use a tool like **guidgen.exe** to generate a valid GUID.  

**Proposed:**
> This value is never saved and only needs to be unique among the other available views for the lookup on the current form. A string for a non-valid GUID will not work, for example “00000000-0000-0000-0000-000000000001” will not work. It’s recommended that you use a tool like **guidgen.exe** to generate a valid GUID.

PR #1626, opened 2020-07-29
Update configure-openid-settings.md

anarchitect:
Because implicit grant is not recommended by OAuth anymore. and some organisation might choose to support Hybrid Flow.  Also, Microsoft is working on support of authorisation flow grant (release date not confirmed yet).  I also added which "site setting" to updated based on experimentation, please confirm this is the correct way of doing it too.


tapanm-MSFT:
Thank you for reaching out and submitting this request. The note is already documented at [Use OAuth 2.0 implicit grant flow within your portal](https://docs.microsoft.com/powerapps/maker/portals/oauth-implicit-grant-flow).

PR #1627, opened 2020-07-29
Update oauth-implicit-grant-flow.md

anarchitect:
Removed Hybrid flow, because I tired /_services/auth/authorize?client_id=[clientid]&redirect_uri=[redirecturl]&response_type=code%20id_token and it returned error which means it doesn't support hybrid flow for this function. 
{"ErrorId":"PortalSTS0007","ErrorMessage":"Value provided for the response type parameter is not a supported value. Please check the value and try again.","Timestamp":"7/29/2020 11:57:31 AM","CorrelationId":"23c83ef5-bea2-4a8c-b864-328523d0248e"}

Also, in the token return, added a clarification that the token returned is an access token (based on OAuth standard and Microsoft convention of access token), this can make this statement a bit clearer.


tapanm-MSFT:
Thank you for reaching out. This will be updated as a part of additional updates to content, hence closing the PR.

PR #1663, opened 2020-08-14
Add reference to async Merge requests

MarkMpn:
Merge requests can also be executed via `ExecuteAsyncRequest`


JimDaly:
@MarkMpn 
I used your XrmToolbox plug-in to run this query:
`select name from sdkmessage where isvalidforexecuteasync = 1`
So I can guess where you are getting your information.

But the message owner needs to approve documenting this.
I've added @mayadumesh to review.

phecke:
@MarkMpn Thanks for the update.

PR #1708, opened 2020-08-29
Update setup-postman-environment.md

JimDaly:
Setting myself as the topic owner


JimDaly:
@phecke 
I've just set myself as the topic owner.

PR #1750, opened 2020-09-18
Adding note regarding several CDSs

BenediktBergmann:
When adding a new App Registration at the moment there are several Common Data Services in the List. In the List the only difference is the GUID that is shown. Its important to choose the correct one.


phecke:
@BenediktBergmann Thanks for suggesting that change.

PR #1763, opened 2020-09-22
Update create-portal.md

carfup:
Adding information regarding the necessary permissions to create Azure AD application while creating a portal


tapanm-MSFT:
Thank you for reaching out. This information is available in the following section:

https://docs.microsoft.com/powerapps/maker/portals/create-portal#disable-portal-creation-in-a-tenant

carfup:
Hello @tapanm-MSFT ,

Thanks for your answer here.
Well in that section, it's written that it's mandatory to have one of those 3 roles but that's not clear.
In my case, my user account on client side is "Power Platform admin" but I still do not have the permission to create a portal...

tapanm-MSFT:
Thanks for the comments. A user with any of the roles [listed](https://docs.microsoft.com/powerapps/maker/portals/create-portal#disable-portal-creation-in-a-tenant) should be able to create a portal. If that doesn't work, lease reach out to support teams for investigation of the issue. Additional information: [Permission matrix](https://docs.microsoft.com/power-platform/admin/use-service-admin-role-manage-tenant#service-administrator-permission-matrix)

PR #1812, opened 2020-10-13
Update openForm.md

ericregnier:
Added formParameters lookup field examples


Nkrb:
@ericregnier if you want to set the lookup field can we just use formParameters["parentustomerid"] = "Id of the record";

ericregnier:
> @ericregnier if you want to set the lookup field can we just use formParameters["parentustomerid"] = "Id of the record";

It might work, but if the field is visible on the form, it will display "(No Name)" as the label.  From what I've tested it works similar to the extraqs in a querystring (https://docs.microsoft.com/en-us/powerapps/developer/model-driven-apps/set-field-values-using-parameters-passed-form#example-set-the-value-for-lookup-fields)

Nkrb (approved):

PR #1942, opened 2020-11-18
Update implementing-controls-using-typescript.md

crowcoder:
There is problem with the operation of the slider when the value is formatted. The range slider value should be set to the raw value converted to a string, not the formatted value. When you set the slider to 1000 it tries to use "1,000" which is not a valid float, and it behave erratically.

[If the value is set to something which can't be converted into a valid floating-point number, validation fails because the input is suffering from a bad input.](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range)

[If parseFloat encounters a character other than a plus sign (+), minus sign (- U+002D HYPHEN-MINUS), numeral (0–9), decimal point (.), or exponent (e or E), it returns the value up to that character, ignoring the invalid character and characters following it.](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseFloat)

### DIFF
--- a/powerapps-docs/developer/component-framework/implementing-controls-using-typescript.md
+++ b/powerapps-docs/developer/component-framework/implementing-controls-using-typescript.md
@@ -168,10 +168,7 @@ export class TSLinearInputComponent
     this._value = context.parameters.sliderValue.raw
       ? context.parameters.sliderValue.raw
       : 0;
-    this.inputElement.value =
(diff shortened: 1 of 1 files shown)


Nkrb:
Closing this PR as we didn't find any relevant information. If you need further help, please reopen or create a new PR.

PR #1949, opened 2020-11-20
Update calculated-rollup-attributes.md

ericregnier:
Updated conflicting limitations:
- Calculated attributes can reference related entities
- Calculated attributes can reference other calculated attributes in their formula, but they can’t reference themselves.


JimDaly:
This is not my topic. Assigning to @mayadumesh

phecke:
@ericregnier Thanks for the doc update.

PR #1999, opened 2020-12-14
Common Data Model seems to have come already

skouti2:


tapanm-MSFT:
Content already updated, closing as orphan.

PR #2010, opened 2020-12-17
Update code samples to work with newer ADAL lib

LordJZ:


phecke:
@LordJZ What version of ADAL are you using with this change?

phecke:
@LordJZ I am closing this PR. I have updated the article to make it current with newer ADAL and MSAL libraries.. Thanks for pointing out that the article needed revision.

PR #2013, opened 2020-12-23
Info on keys and updated database terminology

ChadAlt:
I added text about the need to create alternate keys in the tables in order to set lookup fields or use the "Delete rows that no longer exist..." option. 

All references of entity/entities were changed to table/tables and field/fields to column/columns.


KumarVivek (approved):

phecke:
@ChadAlt Thanks for providing the new information and PR.

PR #2023, opened 2021-01-02
Date column updates

MarkMpn:
Updated description of timezones for date columns to match the current behaviour I see.

It's possible my test instance isn't up to date with the latest updates, but based on my test user settings:

image

I get these results in SSMS:

image

and this in Power BI:

image


phecke:
@MarkMpn This change is planned to be available worldwide by late January or early February. I am going to delay this doc change for a few weeks. Thanks for the PR.

phecke:
@MarkMpn UTC time is now being returned, so I am closing (not merging) this PR as the original documentation text is now correct.

PR #2063, opened 2021-01-19
change 'bar chart' to 'column chart'

Huodong-Yuan:
It is simply column chart, instead of bar chart.


Huodong-Yuan:
Please take a look at this proposed change. If you agree on the change, you may change the name in the path and file name as well.

tapanm-MSFT:
Thank you for reaching out, for more changes will need to update internal repository - your feedback will be reviewed and updated if needed.

PR #2070, opened 2021-01-21
Update use-custom-model-driven-app-on-mobile.md

jeetuGupta:
Added required privileges link for model driven apps.

The missing privilege described here relates to the app being available for mobile. This setting is configured by the app makers, and the sudden change should be related to a change in the app privileges by the maker. 

The documentation that shows what pieces need to be enabled can be found here (Set up Dynamics 365 for phones and tablets | Microsoft Docs).


benabbon (approved):

jeetuGupta:
@mduelae Can you review this please? Let me know if you need anything from my end.

PR #2090, opened 2021-01-29
Update use-change-tracking-synchronize-data-external-systems.md

jlattimer:
Removed errant characters in the url to make it correct


phecke:
@jlattimer Thanks for catching that error in the URL.

PR #2092, opened 2021-01-29
Update best-practices-workflow-processes.md

mariusagur:
Removed real-time from descriptions that that refer to workflows in general
Removed real-time for descriptions that refer to background workflows
Clarified the infinite-loop explanation
Removed "next steps" because those were previous steps, too align with the standard in other docs subsections
Added editor note to consider adding "see also" links to actions and power automate cloud flows.


Mattp123:
@mariusagur - On demand (synchronous) workflows have been re-defined as "real-time workflows" available with Power Apps and background (asynchronous) workflows are maintaining the same name as part of Power Automate. However, your are correct that the "Automatically delete completed real-time workflow jobs" section doesn't belong in this topic because it is referring to background workflows. I've opened and internal task to move that section as well as update the topic to use the new terminology that you mentioned is missing.

Closing without merging.

PR #2118, opened 2021-02-12
Removing outdated common issues - 2182045

ADobin:
A lot of these issues are no longer relevant, this is cleaning up a bunch of old issues to make this list more relevant to users.


tapanm-MSFT:
Closing without merging, updated separately.

PR #2126, opened 2021-02-14
Update handle-exceptions.md

leolidba:
"Initated"  => "Initiated".


JimDaly:
I was just going to fix the typo, but I took the opportunity to refresh the content a bit. 
I hope this is more to the point.

PR #2177, opened 2021-03-09
Standard Table Changes

slalithprasad:
There is no None and Business-owned options for a Custom Standard type Table.


Mattp123:
@slalithprasad - thanks for your feedback. There are system tables that are None or Business Unit owned. From what I could find they are only viewable in solution explorer. There's also a few discrepancies for ownership that's displayed between the Power Apps maker portal and solution explorer that I'm following up internally on. 
I've updated the article to make it more clear about these special ownership types. 
Closing without merging.

PR #2189, opened 2021-03-11
Update control-gallery.md

chmoncay:
Updated text around perferformance


chmoncay:
Updated some text, Milos said the previous wasn't formal enough.

PR #2203, opened 2021-03-16
Update execute-batch-operations-using-web-api.md

JimDaly:
calling out that the operations are performed sequentially in the order they are specified.


JimDaly:
Hi guys. Can you apply this small change?

Nkrb (approved):

PR #2244, opened 2021-04-02
Update delegation-overview.md

kv-prajapati:
Replace a confusing clause with a meaningful one.


tapanm-MSFT:
Will be considered with rest of the reviews and updates for the docs.

PR #2266, opened 2021-04-10
Update use-upsert-insert-update-record.md

JimDaly:
Looks like we got a little carried away the the entity => table renaming in this topic.


JimDaly:
applied fix on internal repo

PR #2285, opened 2021-04-16
Updating to indicate that parameter aliases cannot be used in $expand

bgribaudo:


phecke:
@JimDaly Web API related so assigning to you.

bgribaudo:
Thank you, @JimDaly!

> I think this should work, but the fact that the error code is "code": "0x0",, and the strange nature of the error message suggests this comes from the OData Library.

Any chance that an issue can be opened internally for this to be addressed so that it does work?

JimDaly:
> Any chance that an issue can be opened internally for this to be addressed so that it does work?

I raised this to the API team for awareness but didn't open a bug since the error comes from code we don't own.

bgribaudo:
Thanks, @JimDaly! Hope you have a great Fourth of July!

PR #2312, opened 2021-04-28
Applying Organizational Service remarks to Web API, as well

bgribaudo:
* Adjusting so that remarks about change tracking that currently are worded to just apply to the Organizational Service apply to both that service and the Web API.
* Slight clarification tweaks.


phecke:
Waiting for internal feedback.

JimDaly:
> Does that help?

I don't know that it does. It isn't expected that people manipulate the value, and the date is only part of the token value.
The other part appears to be a version number of some kind.

But I do see that this value returns 30: 
{{webapiurl}}organizations?$select=expirechangetrackingindays

Description for the column is `Maximum number of days to keep change tracking deleted records`.
It seems to be updatable.

I'm looking to talk to the dev team owner about this.

JimDaly:
Reached out to feature team about this again.

bgribaudo:
Thank you!

JimDaly:
Closing this b/c the content today now has the correct expiration day duration of 30 rather than 90.

PR #2355, opened 2021-05-07
Typo fix: Changing "Limits on number of tables -> entities returned"

bgribaudo:


phecke:
@bgribaudo We have to support both Entity Data Model and Power Apps terminology. Thanks for pointing the section title out. I made some adjustments.

PR #2374, opened 2021-05-14
Creation of Sink

pturegano:
The DF UI has changed, this piece of documentation is not current and confussing.


Mattp123:
@pturegano - thanks for your feedback. I'm closing this PR without merging since merging it will not update the article to match the current experience. I've opened a separate internal task to track that work.

Issue #2415, opened 2021-05-29
Reason for recommending Azure B2C over local login


xsatishx:
[Enter feedback here]
Hello - I am new to powerapps and while reading this document I understood that local login is not recommended and instead one has to use Azure B2C login. However, the reason for this recommendation is not very clear in the documentation for someone who is new to PowerApps. 

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 7179b411-acf4-331d-ac80-89a937904a3f
* Version Independent ID: 4dc3d160-c50b-2642-d411-f16e9e056d47
* Content: [Migrate identity providers to Azure AD B2C - Power Apps](https://docs.microsoft.com/en-us/powerapps/maker/portals/configure/migrate-identity-providers)
* Content Source: [powerapps-docs/maker/portals/configure/migrate-identity-providers.md](https://github.com/MicrosoftDocs/powerapps-docs/blob/main/powerapps-docs/maker/portals/configure/migrate-identity-providers.md)
* Service: **powerapps**
* GitHub Login: @sandhangitmsft
* Microsoft Alias: **sandhan**

tapanm-MSFT:
Thank you for reaching out. Local authentication is limited in functionality and may not support enhanced features such as Multi-factor Authentication, Smart Account Lockout and Enforcing Password Complexities.

PR #2419, opened 2021-05-31
Update set-authentication-identity.md

xsatishx:
Why is b2clogin recommended, would it be possible to list down a few reasons? I have customers asking me the reasons for deprecating local login when it works quite well for them and have received feedbacks that the documentation does not give out any valid reason.


tapanm-MSFT:
Please see https://github.com/MicrosoftDocs/powerapps-docs/issues/2415

PR #2424, opened 2021-06-03
Update web-api-perform-operations.md

OOlashyn:
Added important message to notify users that they need to enable collection-valued properties in respective Web API site settings before using them, as from the example it looked like it will work as-is.


tapanm-MSFT:
Thank you for reaching out. This is already mentioned here - https://docs.microsoft.com/powerapps/maker/portals/web-api-overview#site-settings-for-the-web-api starting:
*You must enable the site setting to enable the portals Web API for your portal....*
These settings need to be updated before working on examples. The tutorial is also useful to follow step-by-step: https://docs.microsoft.com/powerapps/maker/portals/web-api-perform-operations#example

PR #2433, opened 2021-06-04
Update set-authentication-identity.md - 2316731

OOlashyn:
Fix for incorrect highlight, missing Microsoft value and non-working new row.


tapanm-MSFT:
Thank you for reaching out. This is being fixed through internal PR for additional fixes and to avoid conflicts.

PR #2441, opened 2021-06-08
Update solutioncomponent.md

ericregnier:
Updated term from CRM to Dataverse


phecke:
@ericregnier The entity reference documentation is generated from the web service. Any change to the published reference docs will be overridden when we next do a doc refresh.

@KumarVivek Can you please verify if Eric's doc change in this PR cannot be done this way.

KumarVivek (approved):

PR #2495, opened 2021-06-28
Update dominant plug-ins description

vashr:


vashr:
Hi @Mattp123 looks like these changes got reverted in master and both the previous description and the new description is showing up.
image

PR #2506, opened 2021-07-04
Update community-resources.md

brasov2de:
Changed to the name of my blog


Nkrb:
@brasov2de PCF is not an approved term to use in the docs, that is the reason we have changed the title. Closing the issue.

PR #2508, opened 2021-07-05
Update work-with-any-data.md

kv-prajapati:
Typo and text formatting


Mattp123:
@kv-prajapati - closing quotation marks outside commas and periods is the correct convention per our style guide. So is the use of contractions. Put those back as per style and accepted other substantive changes.

PR #2518, opened 2021-07-07
Update multitable-lookup.md

bthompson-365:
Removed an errant \ from \@odata.type to @odata.type in the Localized Label part of the Web Api example


phecke:
@bthompson-365 Thanks for noticing that. We have already fixed it.

PR #2522, opened 2021-07-07
Update control-power-bi-tile.md

kv-prajapati:
Explain what if when there are spaces in table and columns.


tapanm-MSFT:
Thank you for reaching out and this change. The documentation will soon have a common change separately for this, so closing this PR.

PR #2562, opened 2021-07-19
Changed to the name of my blog

brasov2de:


Nkrb:
@brasov2de  It is not recommended to use PCF in official docs. If you want to change your blog name to `Dianamics Lady`, we can do that. Please let me know what do you think

brasov2de:
> @brasov2de It is not recommended to use PCF in official docs. If you want to change your blog name to `Dianamics Lady`, we can do that. Please let me know what do you think

I just hoped that "PCFLady" written together, is not against the rules.
With 'Power Apps component framework' inside, is not a name anymore.
Then I would choose 'Dianamics Lady'. Seems to be the lesser evil. 
Thank you!

Nkrb (approved):

PR #2583, opened 2021-07-27
Update Sorting Note

shariquekhan1997:
Changes:
Filtering feature is only available to Dataverse data source. ❌
Sorting feature is only available to Dataverse data source. ✔

Filter gets removed when sorting is re configured.


shariquekhan1997:
In Sorting Note. filtering note is used, in my commit, I corrected it and added one more line which I have tested.

Any update?

Nkrb (approved):

PR #2595, opened 2021-07-29
Clarifying InitializeFrom's behavior: returns a template, does not actually create a new record

bgribaudo:


phecke:
@JimDaly Can you verify this PR and merge. Thanks

JimDaly:
@bgribaudo 
Thanks for your contribution.
Agree that this section needs clarification, but don't want to use the term 'template'.
I've applied some changes in another branch that I hope bring more clarity.  See [Create a new record from another record](https://docs.microsoft.com/en-us/power-apps/developer/data-platform/webapi/create-entity-web-api#bkmk_initializefrom)

bgribaudo:
Thanks, @JimDaly.

PR #2641, opened 2021-08-16
Helpful note to new developers

v22arvind:
If the table columns are not bound to any display control directly, the columns are not retrieved from DataVerse and so they aren't available to use in code too (while writing the code). The solution is to use the  disambiguation operator as suggested in below post:
https://www.itaintboring.com/powerapps/canvas-apps-what-is-a-scope-and-what-does-it-have-to-do-with-filterlookup-functions/


JimDaly:
@tapanm-MSFT Can you review?

KumarVivek:
Closing without merging old items.

PR #2661, opened 2021-08-24
More clarification on setting the Photo property

jimbobbennett:
The **Photo** property is only set when the camera control is tapped or selected. Adding this to clarify as this is something that caused me wasted time trying to work out why I couldn't get an image when using a button, and has caused a number of questions across the internet.


tapanm-MSFT:
Thank you for reaching out and your feedback. This will be updated through internal repo due to additional changes. The new changes will reflect soon.

PR #2728, opened 2021-09-14
web-api-overview.md: "tables" -> "records in tables"

fowl2:
Presumably this was caused by find and replace gone awry.

PS. I think https://github.com/MicrosoftDocs/powerapps-docs/blob/main/powerapps-docs/maker/portals/web-api-perform-operations.md needs to be updated too


tapanm-MSFT:
Thank you for reaching out, this is being updated through internal PR for additional updates. Changes will reflect soon. Closing this PR.

PR #2769, opened 2021-09-29
Add private channel exception

Lennay:


Lennay:
moving this to an diffrent account

PR #2774, opened 2021-10-01
Update function-savedata-loaddata.md

melzoghbi:
Expanding this feature for running apps in web. Latest changes to this feature has been added to support web player and not just in teams or mobile.


melzoghbi:
@gregli-msft can you help reviewing this change our docs? Thanks a lot.

PR #2795, opened 2021-10-11
Adding note about the Default property

pedroangelini:
After a few hours of banging heads against walls, we noticed that the combo box, differently than most controls, does not use the Default property to set the default selected item. Adding a note here in the documentation to prevent more brain trauma.


tapanm-MSFT:
Thank you for your feedback. Documentation has been updated through an internal PR with changes, and the changes will be visible soon. Closing this PR without merging.

PR #2815, opened 2021-10-18
Update function-savedata-loaddata.md - 2519938

melzoghbi:
adding a note about not to store sensitive data in SaveData function in the web


tapanm-MSFT:
Updated using internal PR, closing.

PR #2904, opened 2021-12-06
Update navigate-to-custom-page-examples.md

rfarris2000:
The code sample for AppStart and parsing the recordId is incorrect.  You must use the MID function to strip the value between curly brackets ({GUID})

### DIFF
--- a/powerapps-docs/developer/model-driven-apps/clientapi/navigate-to-custom-page-examples.md
+++ b/powerapps-docs/developer/model-driven-apps/clientapi/navigate-to-custom-page-examples.md
@@ -94,7 +94,7 @@ The `Param` function within the custom page retrieves the value and uses the Loo
 App.OnStart=Set(RecordItem, 
     If(IsBlank(Param("recordId")),
         First(<entity>),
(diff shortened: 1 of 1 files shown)


JimDaly:
@adrianorth 

Second example here [Open as an inline full page with a record context](https://docs.microsoft.com/en-us/powerapps/developer/model-driven-apps/clientapi/navigate-to-custom-page-examples#open-as-an-inline-full-page-with-a-record-context)

Reported: 
> The code sample for AppStart and parsing the recordId is incorrect. You must use the MID function to strip the value between curly brackets ({GUID})

**Current:**
```
App.OnStart=Set(RecordItem, 
    If(IsBlank(Param("recordId")),
        First(<entity>),
        LookUp(<entity>, <entityIdField> = GUID(Param("recordId"))))
    )
```

Proposed:
```
App.OnStart=Set(RecordItem, 
    If(IsBlank(Param("recordId")),
        First(<entity>),
        LookUp(<entity>, <entityIdField> = GUID(Mid(Param("recordId"), 2, 36))))
    )
```

PR #2937, opened 2021-12-25
More info in GitHub PAT

rappen:
Added about Scope=repo - I tested really simple, it should work. For Azure it is easier to create PAT, but GitHub have long different scope, so this info helpes.


tapanm-MSFT:
Thank you for reaching out. This is documented on personal access token topic:

A token with no assigned scopes can only access public information. To use your token to access repositories from the command line, select repo. For more information, see "[Available scopes](https://docs.github.com/en/apps/building-oauth-apps/scopes-for-oauth-apps#available-scopes)".

PR #2973, opened 2022-01-12
Not a preview anymore

brasov2de:
Corresponding to the official blog : https://powerapps.microsoft.com/en-us/blog/custom-page-runtime-has-reached-general-availability/ the Custom Pages (including code components) have reached GA, so it's not a preview feature anymore..


Mattp123:
@brasov2de - you are correct. Removed additional preview tags as well in this pr.

PR #2983, opened 2022-01-14
Update EnvironmentVariables.md

ChrisGarty:
Added a note: Environment variable support for custom connectors is not available yet but is on the roadmap.


ChrisGarty:
Recreating on private...

PR #3022, opened 2022-01-30
Added "old" name Attribute

rappen:
It is used "a lot" by users and developers, so this is found to help users get right to Column.
Just like you already have Entity/Table.


Mattp123:
@rappen - I revised your definition a bit and mentioned the field > column rename mention in the column definition where it really belongs.

rappen:
Thanks @Mattp123 ! 😊
I'm not the best at English these days... 🙄

PR #3025, opened 2022-01-31
IF statements not supported

MarkMpn:
`IF/THEN/ELSE` was listed as supported, but the "Language elements" section also says control-of-flow statements are not supported. Actually running a query using `IF` gives an error, e.g.

```sql
if @@rowcount = 0
begin
  select top 10 name from account
end
```

gives:

```
Msg 40000, Level 16, State 1, Line 15
'IF' statement​ is not supported. Line:1, Position:1
```


phecke:
Waiting for internal feedback.

phecke:
@MarkMpn 

Mark; The documentation is correct. You are combining if and BEGIN which is control of flow. 

For example this statement works. Basically user can only use if within the SQL statement context. Not with statement blocks.

IF DATENAME(weekday, GETDATE()) IN (N'Saturday', N'Sunday')
       SELECT 'Weekend';
ELSE 
       SELECT 'Weekday';

Rich

MarkMpn:
Strangely your example does work, but if I change the SELECT statement to use reference a table I get the original error indicating that IF is not supported:

```sql
IF DATENAME(weekday, GETDATE()) IN (N'Saturday', N'Sunday')
  SELECT top 10 name FROM account;
ELSE
  SELECT top 10 fullname FROM contact;
```

returns:

```
Msg 40000, Level 16, State 1, Line 1
'IF' statement​ is not supported. Line:2, Position:1
RequestId: TDS;d983ad5f-723b-43b5-b141-b90120b1eb89;32
Time: 2022-07-01T19:08:51.2056133Z
```

PR #3065, opened 2022-02-08
Fixing error in Examples & adding Example

BenediktBergmann:
In Example #2 there was an error where type of a simple lookup was set. This would result in JavaScript errors. type is only needed for complex lookups (multi table lookups).


JimDaly:
@adrianorth do you approve?

PR #3090, opened 2022-02-15
Update GetGlobalContext-ClientGlobalContext.js.aspx.md

mspilde:
Removing reference to deprecation of the aspx page support.


KumarVivek:
Addressed vis internal repo. Closing it without merging

PR #3102, opened 2022-02-17
Update 03-building-low-code-prototype.md

mattruma:
Looks like Navigate for OnStart has been deprecated, see http://powerappsguide.com/blog/post/migrate-apps-to-use-app-startscreen, recommended approach is to use StartScreen instead.


spboyer:
@aprildunnam

aprildunnam:
OnStart will continue to work for existing apps but for new apps the setting will be disabled by default to allow Navigate in App.OnStart.  So it's definitely a good idea to update this to use the new StartScreen Property instead.  This looks good to me.

KumarVivek (approved):

KumarVivek:
This change has been done through our internal repo; closing without merging.

PR #3104, opened 2022-02-18
Explicitly state 'X' button returns undefined.

kasmara:
Clicking the 'X' button returns undefined whilst the Cancel button returns the boolean false. 
It would be helpful if this was explicitly stated in the docs rather than having to use guesswork/debugging to discover that.

Open to suggestions about exact wording.


JimDaly:
@adrianorth 

The change to the **successCallback** parameter for the [openConfirmDialog (Client API reference)](https://docs.microsoft.com/en-us/powerapps/developer/model-driven-apps/clientapi/reference/xrm-navigation/openconfirmdialog) seems to make sense.

**OLD:**
 > A function to execute when the confirmation dialog is closed by clicking the confirm, cancel, or **X** in the top-right corner of the dialog. An object with the confirmed (Boolean) attribute is passed that indicates whether the confirm button was clicked to close the dialog.
 
 **PROPOSED:**
> A function to execute when the confirmation dialog is closed by clicking the confirm, cancel, or **X** in the top-right corner of the dialog. An object with the **confirmed** (Boolean) attribute is passed that indicates whether the confirm button (true), cancel button (false) or **X** button (undefined) was clicked to close the dialog.

Need confirmation that this accurately describes the behavior.

PR #3175, opened 2022-03-12
Power Fx (canvas app) code update

last72:
Power Fx (canvas app) code update


powerapps-docs/maker/canvas-apps/add-list-box-drop-down-list-radio-button.md
```diff
@@ -77,8 +77,19 @@ In these steps, you used an expression to create a list of items. You can apply
 
 4. On the **Insert** tab, select **Icons**, and then select the circle.
 
-5. Set the **[Fill](controls/properties-color-border.md)** property of the circle to the following function:  
-   ```If(Choices.Selected.Value = "red", Red, Choices.Selected.Value = "green", Green, Choices.Selected.Value = "blue", Blue)```  
+5. Set the **[Fill](controls/properties-color-border.md)** property of the circle to the following function: 
+
+    ```powerapps-dot
+    Switch(
```
jorisdg, line 83: We recently made a change where enums need to be qualified. So Red won't work anymore you need to specify Color.Red
So any of these identifiers (i.e. when it's not in quotes) should be Color.Green etc.
See https://github.com/MicrosoftDocs/power-platform-pr/pull/6161/files and specifically for colors: https://learn.microsoft.com/en-us/power-platform/power-fx/reference/function-colors

jorisdg (changes requested):
Looks good, but need to fix the colors

tapanm-MSFT:
Closing per comment response and Power Fx updates.

PR #3217, opened 2022-03-24
Add OnPreStageChange to Process events table

rwilson504:
The OnPreStageChange event handler as well as well as its methods were not listed within the 'Process events and event handler methods' table.


JimDaly:
This looks like a miss
This section [Process events and event handler methods](https://docs.microsoft.com/en-us/powerapps/developer/model-driven-apps/clientapi/reference/formcontext-data-process#process-events-and-event-handler-methods)
Doesn't include a reference to [OnPreStageChange event (Client API reference)](https://docs.microsoft.com/en-us/powerapps/developer/model-driven-apps/clientapi/reference/events/onprestagechange)

When it clearly is part of the same group of events.

JimDaly:
@HemantGaur 
Can you approve this change?

JimDaly:
It is already merged.

HemantGaur:
Thanks Jim , this is good approach.


Even I don’t know ownership of certain items beyond formComtext and BPF for my team as I am also new to the setup . Are controls APIs owned by controls team, what about device - mobile and webAPI (infra)? . @Adrian ***@***.***> would it be possible for you to fill in the big buckets with best guess and we can then loop in EMs and PMs from other teams to validate and officially assume the ownership .  Let’s chat tomorrow if you wanna do it together .

Thanks,
Hemant

PR #3230, opened 2022-03-30
Update onreadystatecomplete.md

ericregnier:
Added more details on addOnReadyStateComplete and removeOnReadyStateComplete events and added a sample

### DIFF
--- a/powerapps-docs/developer/model-driven-apps/clientapi/reference/events/onreadystatecomplete.md
+++ b/powerapps-docs/developer/model-driven-apps/clientapi/reference/events/onreadystatecomplete.md
@@ -20,8 +20,17 @@ contributors:
 
 The `OnReadyStateComplete` event indicates that the content of the IFRAME has loaded and can be accessed in code. Use this event when referencing IFRAME controls within your scripts. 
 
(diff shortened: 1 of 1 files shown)


JimDaly:
@ericregnier 
I don't see that we support the `addOnReadyStateComplete` and `removeOnReadyStateComplete` methods for the control.
I didn't find them in the documentation. There are methods that exist, but if we don't document them, they aren't supported.

This shows the methods supported for an IFrame Control:  [iframe control type methods](https://docs.microsoft.com/en-us/powerapps/developer/model-driven-apps/clientapi/reference/controls#iframe-control-type)

Am I missing something?

ericregnier:
Hi @JimDaly,
For iFrames, shouldn't the event be supported, as we can set a handler in the classic UI?
For web resources, the documentation states that _["A web resource control has the same set of methods available as the iframe control. See iframe control type"](https://docs.microsoft.com/en-us/powerapps/developer/model-driven-apps/clientapi/reference/controls#web-resource-control-type)_ and so I assume the same event is supported for web resources. I can confirm that it works (I know it doesn't mean that it's supported thought).

JimDaly:
@sriharibs-msft 
Can you comment on whether we can add Eric's change to the docs?

sriharibs-msft:
This is being driven by @adrianorth and he is best placed to comment. cc @JimDaly

PR #3231, opened 2022-03-30
Update onreadystatecomplete.md

ericregnier:
And Web Resource to title as it applies to Web Resources as well


JimDaly:
Need feature owner to confirm functionality for IFrame also applies to Web Resource.

ericregnier:
Thanks @JimDaly. Happy to jump in a call to discuss more in details with the team if need be.

PR #3252, opened 2022-04-03
Update Visual Studio versions to 2022

Eickhel:
Change from the 2019 version to 2022


JimDaly:
Looking for @noazarur-microsoft  to approve change.
It is good to be up-to-date if VS 2022 is known to work. 
But we could also keep the VS 2019 so people using won't feel that they must upgrade for this.

Eickhel:
Sure @JimDaly, but at least change the Build tools URL because this doesn't work anymore:
https://visualstudio.microsoft.com/downloads/#build-tools-for-visual-studio-2019

JimDaly:
I've applied changes to the main branch to address this. 
Changes should be live by this evening. [Create your first component](https://docs.microsoft.com/en-us/power-apps/developer/component-framework/implementing-controls-using-typescript)
VS 2019 downloads are available here: https://visualstudio.microsoft.com/vs/older-downloads/#visual-studio-2019-and-other-products

PR #3283, opened 2022-04-12
Clarification of duration dimensions

mattgon:
Added row to custom dimension table to describe the ms-duration custom dimension and duration standard dimension


mattgon:
@maustinjones should be a reviewer on this

PR #3287, opened 2022-04-13
AppCenter iOS is called iOS not custom

last72:
AppCenter iOS is called iOS not custom.

See the screenshot below.
image


last72:
Also, button text changed for API token creation (see second commit | 
1e973a5)
image

tapanm-MSFT:
Thank you for reaching out. For iOS, we need custom, hence the documentation is accurate. However, the button text will be updated through internal PR shortly; and the changes should be live soon.

PR #3319, opened 2022-04-25
Add description for Related Records Filtering

Dlutchy:
An explanation for the Related Records Filtering section of the Field Properties is missing. I have added a description.


Mattp123:
@Dlutchy - Related records filtering is considered a special property since it isn't included with many columns so is covered here: [Overview of model-driven app special column properties](https://docs.microsoft.com/en-us/power-apps/maker/model-driven-apps/special-field-properties-legacy)

Closing pull request without merging.

PR #3403, opened 2022-05-26
Update add-custom-javascript.md

oguzerdeve:
The problem with current version, was when user clicks on the validate link,  setfocus is not working. 
I made 2 changes and with this new version scroll and focus is working when user clicks on the validation link (see screen shot added)
I have tested this version and it works as expected.

image


nickdoelman:
code updated in previous update

PR #3440, opened 2022-06-10
Added list with links to properties and methods

beyro:
Every time I come to this page, It's a pain scrolling through to view all the methods and I just want a list at the top I can glance through to see what's available.

### DIFF
--- a/powerapps-docs/developer/model-driven-apps/clientapi/reference/Xrm-Utility/getGlobalContext/userSettings.md
+++ b/powerapps-docs/developer/model-driven-apps/clientapi/reference/Xrm-Utility/getGlobalContext/userSettings.md
@@ -21,9 +21,26 @@ contributors:
 
 Returns information about the current user settings.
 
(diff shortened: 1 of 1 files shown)


JimDaly:
@beyro 
Thank you for contributing, but there is already a list of all the items on the right hand side titled **In this article** that is generated automatically. If we add this hand-edited one we have to maintain it. It duplicates the existing list.

image

Have you used the existing list before?

beyro:
> @beyro Thank you for contributing, but there is already a list of all the items on the right hand side titled **In this article** that is generated automatically. If we add this hand-edited one we have to maintain it. It duplicates the existing list.
> 
> image
> 
> Have you used the existing list before?

No! - I never even noticed that list... but now that you've pointed it out, I will definitely use it :)

JimDaly:
@beyro 
OK. Please try it and I'm going to close this.
Thanks for thinking about ways to improve the docs.

PR #3483, opened 2022-07-02
ContentLanguage property added to button control

last72:
as per referenced here as well https://powerusers.microsoft.com/t5/Building-Power-Apps/ContentLanguage-what-can-we-do-with-it/td-p/1202503

image

image


mduelae:
@clromano - please review.

mduelae (approved):

PR #3490, opened 2022-07-06
Added information on changing current value of variable in managed solution

daniel-amundsen:
We need more information on what to do after a managed solution is imported with environment variables.


Mattp123:
@daniel-amundsen - Although it's true that you would need to add the managed environment variable to an unmanaged solution before you can edit it, that same requirement applies to all managed solution components (as long as the component has the **Allow customizations** managed property enabled). Also, the best practice is to create or use an existing unmanaged solution to make your changes rather than making it in the default solution. See the below note found in this article: [Contextual commands](https://learn.microsoft.com/en-us/power-apps/maker/data-platform/solutions-overview#contextual-commands)

 Note

You can't add components to a managed solution. When you try to, you’ll see the following message:
"You cannot directly edit the components within a managed solution. You’ll need to add it to another unmanaged solution that you’ve created to customize the component. The component might not be customizable."

Closing without merging.

PR #3494, opened 2022-07-07
Added missing word "values"

beyro:


JimDaly:
@beyro 
Thanks for catching this.

PR #3495, opened 2022-07-08
Update configure-entity-relationship-cascading-behavior.md

hambroz:
Fix typos.


phecke:
@hambroz Thanks for catching the spelling errors.

PR #3514, opened 2022-07-19
Changed title to reflect correct function name

JackCerver:
The title 'isOnPremise...' has a missing 's' since the function name is 'isOnPremises'.


JimDaly:
@JackCerver 
Good catch. Thanks for your contribution!

PR #3532, opened 2022-07-28
Undefined variable "privilege" - Update userSettings.md

AshishGupta11011:
Replaced undefined variable "**privilege**" at line 200,201 with "**rolePrivileges**"  to avoid error message "ReferenceError: privilege is not defined"


JimDaly (approved):

JimDaly:
@adrianorth 
I tested this and confirmed if fixes a flaw in the sample code.

PR #3535, opened 2022-07-29
Update dependent-assembly-plugins.md

dimesis:
Fixed misspelling PlugPackage


JimDaly:
@dimesis 
Thanks for fixing this.

PR #3546, opened 2022-08-03
Minor typo in the relationship fix, but crucial

rappen:
I have tried a few time use this how-to-do-it, and I'm pretty sure there is a small "not" in this text.
I may be wrong...

Step 8 in the "Setting up virtual table relationship".


JimDaly:
@rappen

Great timing Jonas. I was just editing this topic in the process of moving it out of the developer content because it doesn't involve writing any code. I'm moving it into the content for makers.

I'm not going to apply this change to this file (because it has moved) but I have applied this change to the file in its new location.

rappen:
Awesome @JimDaly !
You got a link for the new area yet? 😊

JimDaly:
The same link will work. There will be a re-direct in place.

PR #3563, opened 2022-08-10
Add null to return type

beyro:


JimDaly:
@beyro 
Thank you for your contribution!

PR #3567, opened 2022-08-11
documenting that the TDS endpoint requires protocol encryption

dbrownems:
This is from my empirical testing.  You should validate this with the feature owners before merging.


phecke:
@dbrownems Thanks for pointing this out. However, I am closing this PR and adding a link in the topic to existing documentation about the connection. The update will be live in a few hours.

PR #3575, opened 2022-08-16
Adding a know chart limitation

RichardPriddy:
Charts will only render for views that return 50,000 records or fewer.


Mattp123:
@RichardPriddy - This isn't really a limitation as much as it is a design to place limits on reports and views that would cause performance issues. 50K is a large dataset for most applications. I've revised and added a link to the article where this is explained.

PR #3576, opened 2022-08-16
Update return values of getDirection

beyro:
getDirection currently returns "Next" and "Previous" with capital first letters.


JimDaly:
@beyro 
Thank you!

@adrianorth I'm accepting this change based on @beyro assertion.
Let me know if this is not correct.

PR #3583, opened 2022-08-19
Added a new resource

sandeepmishratech:
Hello, I have added a new reference URL for <a> tag in HTML. I came across this tutorial when researching for some materials to help me learn web development. This reference will make the content more valuable. Hope this small contribution will help other learners.


Mattp123:
@sandeepmishratech - Thank you for your feedback. I've replaced your link with a link that points to the W3 org spec for the HTML a element instead of a third-party courseware site.

PR #3585, opened 2022-08-21
Delete duplicated warning

filcole:


JimDaly:
@filcole 
Thanks for finding and fixing this.

Issue #3453, opened 2022-06-17
Info to Set, but not to Get.


devlinl:
This page describes how to configure a form to allow for custom querystring parameters.

However, there does not appear to be any information on how to read the values of those parameters.

After some testing, I discovered that the relevant data is present in Xrm.Utility.getPageContext().input.data["My_Custom_Parameter"]

However, this is not mentioned in the getPageContext() documentation (https://docs.microsoft.com/en-us/power-apps/developer/model-driven-apps/clientapi/reference/xrm-utility/getpagecontext).

Is Xrm.Utility.getPageContext().input.data the correct, supported way to access custom form parameters? 

If so, please update the documentation to explicitly state this.

If not, please create a new page explaining how to access the values of custom form parameters.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 70ab1a46-e673-1783-7bc5-09725a878c4a
* Version Independent ID: 3e54ebd6-296b-815a-e502-cc9e19726013
* Content: [Configure a form to accept custom querystring parameters (model-driven apps) - Power Apps](https://docs.microsoft.com/en-us/power-apps/developer/model-driven-apps/configure-form-accept-custom-querystring-parameters)
* Content Source: [powerapps-docs/developer/model-driven-apps/configure-form-accept-custom-querystring-parameters.md](https://github.com/MicrosoftDocs/powerapps-docs/blob/main/powerapps-docs/developer/model-driven-apps/configure-form-accept-custom-querystring-parameters.md)
* Service: **powerapps**
* Sub-service: **mda-developer**
* GitHub Login: @HemantGaur
* Microsoft Alias: **jdaly**

bishalgoswami:
Thanks for your comment, @devlinl. This issue has been assigned to the doc author for review.

JimDaly:
@devlinl 
Thanks for your feedback.

@HemantGaur to confirm, but I believe this is by design.

The fact that `Xrm.Utility.getPageContext().input.data` is not documented means it is not supported.

The capability to set custom query string parameters is intended for the purpose mentioned in the doc:

> ...model-driven apps allows a specified set of query string parameters to be passed to a form. You use these parameters to set default values when you create a new record in the application. 

The values in the query string parameters should be available as the values of those columns in the form. 
I would expect them to be included within `formContext.data.entity`. Why not access them there?

How would accessing the data via `input.data` property add value?

[...]

Issue #2832, opened 2021-10-26
Accessing formParameters passed from openForm


fowl2:
This article is linked to as reference for the `formParameters` parameter of [`Xrm.Navigation.openForm`](https://docs.microsoft.com/en-us/powerapps/developer/model-driven-apps/clientapi/reference/xrm-navigation/openform), but it doesn't actually say how to access those parameters. These at least appear to be available as `input.data`, but I can't find this documented anywhere.

If `input.data` is supported, I'd be more than happy to create a PR updating the docs :), otherwise, how are parameters accessed? 

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 70ab1a46-e673-1783-7bc5-09725a878c4a
* Version Independent ID: 3e54ebd6-296b-815a-e502-cc9e19726013
* Content: [Configure a form to accept custom querystring parameters (model-driven apps) - Power Apps](https://docs.microsoft.com/en-us/powerapps/developer/model-driven-apps/configure-form-accept-custom-querystring-parameters)
* Content Source: [powerapps-docs/developer/model-driven-apps/configure-form-accept-custom-querystring-parameters.md](https://github.com/MicrosoftDocs/powerapps-docs/blob/main/powerapps-docs/developer/model-driven-apps/configure-form-accept-custom-querystring-parameters.md)
* Service: **powerapps**
* Sub-service: **mda-developer**
* GitHub Login: @Nkrb
* Microsoft Alias: **nabuthuk**

JimDaly:
I think this was addressed in [#3453](https://github.com/MicrosoftDocs/powerapps-docs/issues/3453)
No documentation about this because it isn't supported.

PR #3595, opened 2022-08-24
Update description of data attribute for `navigateTo` method

beyro:
* Provided more details about the data attribute when navigating to a form - especially that custom parameters are accessed via `formContext.data.attributes` because I can't find this information anywhere else in the docs and I always end up in a circular search for 30 minutes before I figure out how it works again.


JimDaly:
@HemantGaur please review this.
Also see issues #2832 & #3453 which seem to be related.

JimDaly (approved):

JimDaly:
@beyro 
Thank you for your contribution!

PR #3661, opened 2022-09-19
Fix hyperlink for onOutputChange

filcole:
Hyperlink pointing to onselection event instead of onoutputchange


JimDaly:
@filcole 
Thanks for catching this.

PR #3666, opened 2022-09-20
Changing getOutput to getOutputs

melamriD365:
getOutput method is not defined. We should use getOutputs method


JimDaly:
@melamriD365 
Thanks for catching this.
There are other topics that have links to this, and I'm re-naming the file, so applying changes in a different branch.
Should be fixed by tonight.

PR #3671, opened 2022-09-21
Fix typos in "Customize the editable grid control"

LesserCure:
- Faciltate -> Facilitate
- Principal -> Principle


JimDaly:
@LesserCure 
Thanks for catching these.

PR #3728, opened 2022-10-09
Selected property of Combo Box Control

poweractivate:
Selected property of Combo Box Control is not present in the docs, but it is a valid property.
 This proposed change adds the following description to Key properties of Combo Box.
**Selected** – The last selected item resulting from user interaction.


poweractivate:
Before merging this pull request, please have appropriate team check the discussion here in detail:
https://powerusers.microsoft.com/t5/Building-Power-Apps/ComboBox-IsEmpty-not-working/td-p/1807664
I suspected that the Selected property on a Combo Box may have been intentionally missing.
However, it's possible, it's missing by mistake.
Which is it?
If it's missing by mistake, please consider merging my pull request, or adding something similar yourself along these lines.
If it's missing intentionally, please reject my pull request. In that case, I might like to know if it is a deprecated property, and if so, I might consider to submit another pull request indicating that the property is deprecated, like is currently the case of the Default property.

mduelae:
@clromano - please review

clromano (approved):

PR #3796, opened 2022-11-03
replaced semicolon with comma

rafche:


Mattp123:
@rafche - Since the semicolon is separating two complete clauses in the sentence either a comma or a semicolon is suitable in this context. Closing without merging.

PR #3812, opened 2022-11-09
Update define-custom-state-model-transitions.md

ericregnier:


JimDaly:
Thanks for your contribution. This is good information that is missing in our content today.
The only problem is that this topic (`define-custom-state-model-transitions.md`) is about defining custom state model transitions, which is an option when configuring definitions for incident and custom tables.

The point you are raising is bigger than this. You are addressing a gap where there isn't any content that addresses working with state and statuscode values on any table, regardless of whether custom state model transitions are applied.

So, where to include this? I think this is something people should be aware of whenever performing an Update. It also needs to be applied for both SDK and Web API.

Today we have two different topics for Update:

- [Update and delete table rows using the Organization Service](https://learn.microsoft.com/en-us/power-apps/developer/data-platform/org-service/entity-operations-update-delete)
- [Update and delete table rows using the Web API](https://learn.microsoft.com/en-us/power-apps/developer/data-platform/webapi/update-delete-entities-using-web-api)

Long term, we have a plan to unify these two paths to describe behaviors (like this) that are common in Dataverse regardless of the programming style. But in the short term, it is acceptable add this to the existing topics that discuss update operations.

Thoughts?

ericregnier:
Thanks @JimDaly. It totally makes sense and I have updated as advised. See PRs:
https://github.com/MicrosoftDocs/powerapps-docs/pull/3824
https://github.com/MicrosoftDocs/powerapps-docs/pull/3825

JimDaly:
Closing in favor of other PRs 3824 and 3825

PR #3824, opened 2022-11-14
Update entity-operations-update-delete.md

ericregnier:

### DIFF
--- a/powerapps-docs/developer/data-platform/org-service/entity-operations-update-delete.md
+++ b/powerapps-docs/developer/data-platform/org-service/entity-operations-update-delete.md
@@ -29,6 +29,16 @@ Each of the examples uses a `svc` variable that represents an instance of a clas
 >
 > You should create a new `Entity` instance, set the id attribute and any attribute values you are changing, and use that entity instance to update the table row.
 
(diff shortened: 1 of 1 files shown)


JimDaly:
@ericregnier 
Thanks for your contributions.
This topic has a lot of important notes, so I'm not too keen on adding another.
There is some other clean-up I want to do in this topic, so I won't accept this 'as-is', but I will leave it open until I make the changes.

It is a pretty big gap that should be mentioned. Thanks for raising it.

ericregnier:
My pleasure @JimDaly. Maybe we can restructure the important notes and create seperate sections for each "important note" part? The statecode and statusreason behavior is really important and impact can be huge so keen to have this clarified in the docs. Cheers!

JimDaly:
@ericregnier 

For this, I added a note to some common content we have about the behavior of statecode and statuscode fields. 

It will appear in two places where this common content appears:

- https://learn.microsoft.com/en-us/power-apps/developer/data-platform/special-update-operation-behavior
- https://learn.microsoft.com/en-us/power-apps/developer/data-platform/org-service/entity-operations-update-delete#legacy-update-messages

PR #3825, opened 2022-11-14
Update update-delete-entities-using-web-api.md

ericregnier:


JimDaly:
@ericregnier 

I added a note to this topic. Re-worded and not with the example you provided, but I hope it gets the point across.

PR #3889, opened 2022-12-08
Asynchronous plug-in steps and workflows

n0rthface:
We discovered recently that asynchronous plug-in steps and workflows are not runnning anymore when bypass custom plugins is enabled for requests. We did test it back and forth and could verify this behavior consistently.


JimDaly:
@n0rthface 
Thanks for the feedback.
The documentation describes the correct behavior of this parameter. I am aware that there is a bug that causes what you see.
I recommend that you contact technical support to report that this is impacting you.

PR #3905, opened 2022-12-15
Update implementing-controls-using-typescript.md

JimDaly:
Moved `npm run refreshTypes` command step to after editing the manifest file.


JimDaly:
Will fix via other means

PR #3917, opened 2022-12-22
Suggested changes to reference guidance

HenryJammes:
Suggesting to reference the Power BI Guidance document into relevant pages on the Power Platform site to aid in discoverability.


Mattp123:
@HenryJammes - Since this PR is in the public repo and out of date, I've manually re-created your PR in the private repo. It will be published live in the next few hours.

PR #3919, opened 2022-12-23
Update code-components-best-practices.md

miguelverweij:
The linting is explained and useful for a tutorial, but it gives some errors because the react version is not specified. I've added this step.

image


JimDaly:
@kaushikkaul 
Can you review please?

PR #3930, opened 2023-01-05
Updated Test Studio known limitation

ChrisBal-MSFT:
Adding a known limitation bullet point to note that Test Studio does not work with Git Version Control.  This is a common issues that is talked about on the Git Version Control doc page, but has not been mentioned on the Test Studio page.


tapanm-MSFT:
Closing since updated with internal repo.

PR #3932, opened 2023-01-06
Update ribbons-available.md Form ribbons

brflood:
Add note indicating that Form ribbons is not supported in UCI

https://dynamicscrm.visualstudio.com/OneCRM/_workitems/edit/2937758


JimDaly (changes requested):
Please move note directly below the section heading.

JimDaly (approved):

PR #3950, opened 2023-01-15
Clarified Default property limitation

BCBuizer:
Moved explanation of the limitation to a single default item from the Selected property description the Default property description.


mduelae:
@closing this since similar updates were made in the internal branch.

Issue #3974, opened 2023-01-24
Link "Configure tables and attributes for auditing" leads to page not found


kannan-balasubramanian:
[Enter feedback here]
The link with the text "[Configure tables and attributes for auditing](https://learn.microsoft.com/en-us/power-apps/developer/data-platform/configure-entities-attributes-auditing)" leads to a 404 page not found.

---
#### Document Details

⚠ *Do not edit this section. It is required for learn.microsoft.com ➟ GitHub issue linking.*

* ID: a5c391f2-d677-6d8b-6bfc-aac9796670f4
* Version Independent ID: 09fe5f18-2a69-2c05-2a86-5b243a75bd9e
* Content: [Work with any data - Power Apps](https://learn.microsoft.com/en-us/power-apps/maker/data-platform/work-with-any-data)
* Content Source: [powerapps-docs/maker/data-platform/work-with-any-data.md](https://github.com/MicrosoftDocs/powerapps-docs/blob/main/powerapps-docs/maker/data-platform/work-with-any-data.md)
* Service: **powerapps**
* Sub-service: **dataverse-maker**
* GitHub Login: @matp
* Microsoft Alias: **matp**

Mattp123:
@kannan-balasubramanian - Thank you for pointing this out. The link has been updated to point to the correct article on configuring auditing for Dataverse table records. 
Closing issue.

PR #4027, opened 2023-02-07
Fixed link

BCBuizer:
Fixed link for Configure tables and attributes for auditing as per #3974


Mattp123:
Revised link to go to top of article instead of a subsection.

PR #4072, opened 2023-02-21
Update component-framework-for-canvas-apps.md

mabrizi01:
in the first paragraph, I reversed the order of App Maker and Professional Developer tasks. It should be the Professional Developers who develop the PCFs, while the Makers import them into their applications


JimDaly:
@mabrizi01 
Good catch. Thanks for fixing.

PR #4092, opened 2023-02-23
Line 178 : Mistake

AshishGupta11011:
Plugin should take max 2 minutes instead of 2 seconds execution time.


phecke:
@AshishGupta11011 Thanks for the feedback. This PR is no longer needed as the text in question was removed, and mention of the 2-minute limit moved to the plug-in performance topic.

JimDaly:
@phecke  & @AshishGupta11011 

That was not a mistake.  It wasn't intended to communicate the limit. It was intended to set the expectation that **_synchronous_** logic that requires more than _**2 seconds**_ is going to contribute to a poor user experience.

While there is a 2-minute time out, it is extremely generous, and people should not consider this the limit they must stay below.

If that wasn't clear, perhaps we can find a way to express it differently.

AshishGupta11011:
Okay , Thanks @JimDaly & @phecke .

PR #4098, opened 2023-02-24
Update create-package-app-appsource.md

hassansuhaib:
Added a missing left square bracket.


phecke:
@hassansuhaib Thanks for the text correction.

PR #4134, opened 2023-03-03
Update performance-tips.md

MichelleMSFT:
- Moved "Avoid control dependencies between screens" above "limit the number of controls". These two topics are both referencing controls so they should be near one another for context. 

Updated control performance recommendations: 
-- Canvas Apps now triggers the warning for 300 controls per screen instead of 500.  
-- Changed text to refer to control limits per screen instead of per app, due to the rendering of screens. 

Updated control dependencies between screens
- Updated a note that there is no had rule on # of screens, but the recommended step if Power Apps studio begins to slow down.


mduelae:
outdated - content updated by pm team

PR #4143, opened 2023-03-07
Update use-fetchxml-construct-query.md

fordosa90:
Add remarks regarding XML encoding in the provided example


JimDaly:
@fordosa90 
Andras
You are including content that says 

> always make sure that your string is a valid encoded XML string

What is the issue you are trying to guide people to avoid? What kind of error would occur if the XML string isn't encoded as you recommend?

This topic is about working with FetchXml with either the SDK or the Web API.

With Web API, the fetchxml must be URL encoded. Not sure that is necessary when using the SDK.

So, perhaps this topic is the one that you want to edit? [Use FetchXML with Web API](https://learn.microsoft.com/en-us/power-apps/developer/data-platform/webapi/use-fetchxml-web-api)

fordosa90:
hey @JimDaly, this is a real quick reply!

it is relevant for the provided example using the SDK, passing in the xml as a string to the fetch

if your constructed string - for example in the code snippet above - contains invalid XML characters, such as & or <, namely instead of "Cannon" it would be "Cannon&Shaw", the snippet runs with InvalidXml error. On last names is not truly relevant, yet, the above snippet shows a generic usage of fetchXml with SDK through a constructed string, so worth the mention, no?

JimDaly:
@fordosa90 
BTW, thanks for your contribution.

Ah, I see. The guidance here seems to be specifically about the `condition` element `value` attribute.  I was confused by other requirements to encode the entire XML string when using Web API.

What you are saying sounds right. We should avoid the error and make sure that the matching criteria will be returned.

An internet search led me to this blog post [Returning an XML Encoded String in .NET](https://weblog.west-wind.com/posts/2018/Nov/30/Returning-an-XML-Encoded-String-in-NET)

So, perhaps the example we have today could be replaced with something like this:

```
string fetch2 = $@"  
    <fetch mapping='logical'>  
        <entity name='account'>   
        <attribute name='accountid'/>   
        <attribute name='name'/>   
        <link-entity name='systemuser' to='owninguser'>   
            <filter type='and'>   
                <condition 
                attribute='lastname' 
                operator='ne' 
                {new XAttribute("value", "Cannon & Shaw")} />   
            </filter>   
        </link-entity>   
        </entity>   
    </fetch>";
```
But I see that the output has double quotes around that attribute:

```
    <fetch mapping='logical'>
        <entity name='account'>
        <attribute name='accountid'/>
        <attribute name='name'/>
        <link-entity name='systemuser' to='owninguser'>
            <filter type='and'>
                <condition
                attribute='lastname'
                operator='ne'
                value="Cannon &amp; Shaw" />
            </filter>
        </link-entity>
        </entity>
    </fetch>
```

Not sure if that will be a problem or not. Couldn't find a way to change it.

Is there a particular method you have used?

fordosa90:
Happy to contribute, this sub page seem to be extremely quick in responses!

No particular method for me, depends always on the end purpose (speed, limitations, ..)
The mentioned blog does discuss nicely some potential steps

I would say it is sufficient to leave the example snippet as is, given it does provide an easily digestible example for most, as in usual scenarios you actually won't see issue - not many text fields contain &<>'" in a Dataverse data model and used in custom queries...

My thought process was, that it is nice to have the mention in that note/important box for those who run into the issue to encode the value based on their choice

JimDaly:
@fordosa90 
Thanks, so this would be a simpler way to address it:
```
<filter type='and'>   
    <condition 
    attribute='lastname' 
    operator='ne' 
    value='{SecurityElement.Escape("Cannon & Shaw")}' /> 
```

When possible, I like to ask folks questions about their submissions ASAP. After too much time passes, people may have forgotten why they thought the issue was important. In this case, I was confused and needed clarity. Thanks.

This is assigned to @phecke, so I'll let him merge it.

phecke (approved):

phecke:
@fordosa90 Thanks for your contribution to our docs.

PR #4232, opened 2023-03-27
Update getActiveProcess.md

Anwarluck:
On line 36 there was an extra bracket after process '[setActiveProcess)]', which I thought came as a mistake, please review.


JimDaly:
@Anwarluck 
Thank you for helping make our docs better!

PR #4264, opened 2023-04-04
Update Limitations

TKlingebiel:


TKlingebiel:

JimDaly:
I'm not sure that this kind of filter requires the `value-of` attribute that was added for column comparison.
I'm sure it works, but could it be done without it?

TKlingebiel:
> I'm not sure that this kind of filter requires the `value-of` attribute that was added for column comparison. I'm sure it works, but could it be done without it?

Not really. Without comparing between table columns or subquery as in sql some tasks required postprocessing fo the data after queryexecution.

PR #4273, opened 2023-04-05
Update getActiveStage.md

Anwarluck:
On line 36 there was an extra bracket after process '[setActiveStage)]', which I thought came as a mistake, please review.


JimDaly:
@Anwarluck 
Thanks for helping make our docs better.

PR #4295, opened 2023-04-12
Update app-visibility-privileges.md

Anshulmsft:
User must have both create and write privilege to access all apps


KumarVivek:
@Anshulmsft Closing this without merging as we have taken care of this internally.

PR #4352, opened 2023-05-02
Update common-field-properties-legacy.md

igorantunesms:
Updating the Searchable property to reflect that it only affects "Edit Filters" but not "Edit Columns" in UCI, which is the intended behavior per PM.


Mattp123:
@igorantunesms - What fields aren't displayed when the Searchable option isn't selected?

PR #4363, opened 2023-05-04
Use System.Configuration.ConfigurationManager NuGet package

richardcarrigan:
The System.Configuration reference is no longer available. It has been replaced by the System.Configuration.ConfigurationManager NuGet package, so I've updated the instructions to match, including a screenshot similar to the one used in the quick-start guide.


richardcarrigan:
> @microsoft-github-policy-service agree

JimDaly (approved):
@richardcarrigan 

Thanks for your contribution.

There are several other things wrong with this sample, but this is a good start.

richardcarrigan:
> @richardcarrigan
> 
> Thanks for your contribution.
> 
> There are several other things wrong with this sample, but this is a good start.

Agreed, @JimDaly . I was able to successfully get this to build today with MFA instead of including the username and password, which was throwing errors. Should I submit a PR reflecting that change, or would that be too opinionated for a sample like this?

Also, on a somewhat related note, I rebuilt this same app using VS Code and the .NET CLI. Is there a way for me to contribute that to this doc, or is there a better way to add that documentation?

JimDaly:
@richardcarrigan 

After I approved your PR, I went in and make a few more changes to reflect my experience while completing the exercise. But these are just patches until a larger change can be made.

The purpose of this sample is to:

 - Demonstrate that people can use their own coding patterns and that the initial quick start is a minimum
 - Guide them towards better practices that are demonstrated in our other C# Web API samples.

But since this was originally written, our other C# Web API samples were updated to use the [WebAPIService class library](https://learn.microsoft.com/en-us/power-apps/developer/data-platform/webapi/samples/webapiservice) which is used by all our newest C# Web  API samples in [PowerApps-Samples/dataverse/webapi/C#-NETx/](https://github.com/microsoft/PowerApps-Samples/tree/master/dataverse/webapi/C%23-NETx)

It has been my intention to re-write this enhanced quick start as an introduction to the WebAPIService example.

WebAPIService is just a sample, but it demonstrates some key design points that are important to be productive using all parts of the Web API in a consistent manner, especially including capabilities such as $batch, managing errors for service protection limits, and guidance that emerged about using HttpClient correctly. (See [Use IHttpClientFactory to implement resilient HTTP requests](https://learn.microsoft.com/en-us/dotnet/architecture/microservices/implement-resilient-applications/use-httpclientfactory-to-implement-resilient-http-requests), and blog posts such as [You're (probably still) using HttpClient wrong and it is destabilizing your software](https://josef.codes/you-are-probably-still-using-httpclient-wrong-and-it-is-destabilizing-your-software/), etc)

Just sharing because that represents the direction I was planning to go with this, but I haven't found the time yet to update this important article.

Moving the content 'as-is' today to use VSCode and .NET CLI isn't a priority since the current content doesn't move towards the goals I've had in mind. The current design of adding helper methods to a partial program class is pretty primitive. I think people prefer to use appsettings.json rather than App.Config for example. Getting away from password in the connection string too, as you mentioned.

But I would very much welcome a way to help people transition past the quick start into the next level, even if it isn't using WebAPIService. I don't want to be proscriptive that people should model their code after WebAPIService because that is just the pattern we have chosen to follow for our .NET Console application sample code. It isn't an SDK.

I hope this makes sense.

The current [EnhancedQuickStart](https://github.com/microsoft/PowerApps-Samples/tree/master/dataverse/webapi/C%23/EnhancedQuickStart) is .NET Framework. 
I'd like to retire this completely.

We don't have an EnhancedQuickStart in our [C#-NETx](https://github.com/microsoft/PowerApps-Samples/tree/master/dataverse/webapi/C%23-NETx) which is where we need the .NET Core version.

If you want to fork the [microsoft/PowerApps-Samples](https://github.com/microsoft/PowerApps-Samples) and include a PR with some of your ideas, that could be a good way to proceed.

richardcarrigan:
Thank you, @JimDaly , for taking the time to provide this insightful information. I'm currently reading/working through the Dataverse Developer guide, which has a "side quest" into both the Web API and the Organization Service, so I'm probably not best suited to help with this refactor... yet 😁.

That being said, I'm definitely interested in contributing to the MSFT docs in a larger capacity, since I refer to them constantly, so I'll keep this in mind to come back to when I feel like I have a better grasp on how *best* to interact with the Dataverse through code.

Till next time!

PR #4409, opened 2023-05-28
Update low-code-plug-ins.md

nathanroseakl:
1/ instant can't be called from modern commanding
2/ calling external service can only be done via instant plugin 3/ external service appears to only support sql server


Mattp123:
@nathanroseakl - There are a couple of issues with your PR.

1. Microsoft employees use the private repo to submit there contributions.
2. Your PR is based on a forked branch. When creating your PR in the private repo, use a branch directly off of the main branch. 

Your PR is being closed without merging. I'll follow up directly with you and provide resources where you can get information about how to contribute in Power Platform repos.

PR #4412, opened 2023-05-30
Update property.md

brasov2de:
property usage can be of type input, bound or output


JimDaly:
@brasov2de 
Thank you for your contribution.

I confirmed that we have published samples using `usage="output"`

PR #4426, opened 2023-06-04
Update object-output.md

x-delfino:
Updated object-output.md to correct code for `getOutputSchema` method as per the [ObjectOutputControl Sample](https://github.com/microsoft/PowerApps-Samples/blob/master/component-framework/ObjectOutputControl/ObjectOutputControl/index.ts#LL51C1-L55C6).

Looks like the previous code block was duplicated by mistake


x-delfino:

JimDaly:
@x-delfino 

Thanks for finding and correcting this.

PR #4481, opened 2023-06-23
Update modern-control-dropdown.md

jblfan79:
Add an explanation for the DefaultSelectedItems field of the control. Also added an example for users who would use a choice column from a sharepoint list


mduelae:
old/outdated - content already exists.

PR #4486, opened 2023-06-26
Update override-default-open-behavior-grids.md

ericregnier:
Add a new updated approach to modify the default open behavior with the new Power Apps Grid Control


JimDaly:
@ericregnier 
Thanks for your contribution.

This change is to modify this article: [Override the default open behavior of data rows in an entity-bound grid](https://learn.microsoft.com/en-us/power-apps/developer/model-driven-apps/override-default-open-behavior-grids)

I reviewed this but didn't test it end-to-end. How would you characterize the difference between this new approach from the existing one?

As far as I can see, the new approach leverages the fact that the Power Apps Editable grid control exposes the `OnRecordSelect` event.

If I understand the instructions, the new approach would *immediately* open the new URL when the record is selected. Whereas if I understand the existing approach, the change enables a new custom button that can be used to open a new window when the record is selected.

Does this match your understanding?

Here is a problem I see:

This article: [Override the default open behavior of data rows in an entity-bound grid](https://learn.microsoft.com/en-us/power-apps/developer/model-driven-apps/override-default-open-behavior-grids) is about customizing the command bar using "classic commands".

But your approach doesn't use the command bar or any command.. so while it achieves the same or similar outcome, I think it needs to be in a different location within the docs.

It seems to have more in common with this article [Navigating to a custom page using client API](https://learn.microsoft.com/en-us/power-apps/developer/model-driven-apps/clientapi/navigate-to-custom-page-examples).
If I'm not mistaken, it is similar to the existing example: [Open from a selected record in editable grid as a centered dialog with record ID](https://learn.microsoft.com/en-us/power-apps/developer/model-driven-apps/clientapi/navigate-to-custom-page-examples#open-from-a-selected-record-in-editable-grid-as-a-centered-dialog-with-record-id) except it uses `Xrm.Navigation.openUrl` rather than `Xrm.Navigation.navigateto`.

Does this seem accurate to you?

ericregnier:
Thanks for your prompt comments @JimDaly!
If I understood your question correctly, it's a different approach because it's a different way of modifying the default behavior open record on model-driven app grids - with the client-side event instead of a custom ribbon command. In my opinion, the new approach is way simpler for the customizers.
From the user/developer/customizer perspective, I suggest keeping it here because it's easier to find and compare the desired approach. I see the article [Override the default open behavior of data rows in an entity-bound grid](https://learn.microsoft.com/en-us/power-apps/developer/model-driven-apps/override-default-open-behavior-grids) about customizing the default behavior whether it's with the ribbon or not... And most likely the users will search how to modify the record select/open behavior and can see all the options in one place. Perhaps in the future there will be other ways (hopefully without even requiring code) and can be updated here.

Hope this makes sense?

JimDaly:
> bout customizing the default behavior whether it's with the ribbon or not... And most likely the users will search how to modify the record select/open behavior and can see all the options in one place. Perhaps in the future there will be other ways (hopefully without even requiring code) and can be updated here.

@ericregnier 

I hear what you are saying. There can be different ways to do the same thing. That makes things challenging because our content is organized by feature and this article (_and all the ones it is grouped with_) is about customizing ribbon commands. 

image

Each article in this section calls out that:

> This topic is about classic commands.
> 
> There is a new way to define commands. See [Modern commanding overview (preview)](https://learn.microsoft.com/en-us/power-apps/maker/model-driven-apps/command-designer-overview).

At some point in the future when modern commanding replaces ribbon, all this content will be removed. No one will look into this article to extract the separate approach that doesn't use ribbons to achieve the same result. It would get lost.

I would prefer to have a separate page that highlights scenarios around using `Xrm.Navigation.openUrl`, where this particular scenario could be used. And then we could have links between these two articles to highlight that they achieve the same result.

What do you think?

JimDaly:
Or we could put this as an example for the [Grid OnRecordSelect event ](https://learn.microsoft.com/en-us/power-apps/developer/model-driven-apps/clientapi/reference/events/grid-onrecordselect) page.

Lots of room on that page...

ericregnier:
@JimDaly yup that makes sense based on how the info is structured now. I will move it to https://learn.microsoft.com/en-us/power-apps/developer/model-driven-apps/clientapi/reference/events/grid-onrecordselect as suggested and create a new PR

ericregnier:
https://github.com/MicrosoftDocs/powerapps-docs/pull/4492

PR #4490, opened 2023-06-26
Update transform-infopath.md

swatantra:
Power Apps is a PaaS, not the SaaS as mentioned.


mduelae:
topic no longer exists

Issue #4517, opened 2023-07-04
Attributes OnChange and OnSelect Missing for Tabs


divya-akula:
[Enter feedback here]

---
#### Document Details

⚠ *Do not edit this section. It is required for learn.microsoft.com ➟ GitHub issue linking.*

* ID: 52f4635c-5645-1042-20ce-dc99498e4a8b
* Version Independent ID: 52f4635c-5645-1042-20ce-dc99498e4a8b
* Content: [Tabs or tab list modern control in Power Apps - Power Apps](https://learn.microsoft.com/en-us/power-apps/maker/canvas-apps/controls/modern-controls/modern-control-tabs-or-tabs-list)
* Content Source: [powerapps-docs/maker/canvas-apps/controls/modern-controls/modern-control-tabs-or-tabs-list.md](https://github.com/MicrosoftDocs/powerapps-docs/blob/main/powerapps-docs/maker/canvas-apps/controls/modern-controls/modern-control-tabs-or-tabs-list.md)
* Service: **powerapps**
* Sub-service: **canvas-maker**
* GitHub Login: @yogeshgupta698
* Microsoft Alias: **yogupt**

bishalgoswami:
Thanks for your comment, @divya-akula. This issue has been assigned to the doc author for review.

divya-akula:
@bishalgoswami  Fixed and created a pull request for the same [https://github.com/MicrosoftDocs/powerapps-docs/pull/4518] (Pull Request)

PR #4518, opened 2023-07-04
Attributes onChange and OnSelect is missing #4517

divya-akula:
Updated the documentation for the missing properties


mduelae:
old/outdated - content is already updated

PR #4523, opened 2023-07-06
Update long-term-retention.md

pnghub:
Added the below under Important:
> - Customers that bring own key(BYOK), note that long term retained data in the Dataverse managed data lake is encrypted with Microsoft managed key only.

Made some changes:
*Long-term retention* (LTR) is a capability that enables customers to transfer their data from a Dataverse transactional database to a managed data lake. 

TO 

*Dataverse long-term retention* (LTR) is a capability that allows customers to securely store their in-active rarely accessed data, in a Dataverse managed data lake.


pnghub:
Need to rollback as the changes were not made to review.learn site.

PR #4674, opened 2023-08-24
Clarified maximum on NumberOfSeries property

BCBuizer:
There is a maximum of 9 series that can be displayed for this control, but that limitation is not documented, hence I added it.


mduelae:
@clromano please review and approve

PR #4719, opened 2023-09-11
Update create-and-use-dataflows.md

ananth11:
Add note about DLP policy for Web Contents blocking Sharepoint too


KumarVivek:
Closing without merging and using internal repo instead for this change.

PR #4773, opened 2023-09-28
Update file-column-data.md

MuhamadUsmankhan:
There were some wording and grammar mistakes I think that I removed.


JimDaly:
@MuhamadUsmankhan 
Thanks for catching these!

PR #4788, opened 2023-10-02
Update events.md

nstratton-srx:
Grammar edit.


JimDaly:
@nstratton-srx 
Thank you for helping make our docs better.

PR #4898, opened 2023-11-03
Update azure-synapse-link-delta-lake.md

jminciper:
Added a note around recommended spark cluster configuration to advise users on lower tier subscriptions that their subscription may need a vCore quota increase in order to successfully set up Synapse Link in the Delta Lake format.


jminciper:
@JasonHQX , what is the typical SLA for these sorts of merges to be reviewed and merged (if applicable?)

jminciper:
@KumarVivek @Mattp123, is this ever likely to be merged?

PR #4899, opened 2023-11-03
Update add-ai-chatbot.md limitations

lindaluomsft:
Adding limitation for the chatbot control specifying that it does not support connecting to regional PVA bots


mduelae:
updates made here: https://github.com/MicrosoftDocs/powerapps-docs-pr/pull/9059

PR #4958, opened 2023-11-19
Update powerapps-overview.md

shaheerahmadch:
Added :

Get Developer Tenant
If you don't have an organization account or aren't part of a tenant or organization account, you can establish a developer's account through the [Microsoft Developer Program](https://developer.microsoft.com/en-us/microsoft-365/dev-program).


mduelae:
info coved in try it for fee section

PR #4961, opened 2023-11-20
Update addPreSearch.md

GuidoPreite:
fixed typo


JimDaly (approved):
@GuidoPreite 
Thanks for fixing these.

PR #4994, opened 2023-11-29
Inserted space in PowerApps

BCBuizer:
Inserted space in PowerApps


JimDaly (changes requested):
The link text PowerApps should be changed to `Power Apps`, but there are other changes here to the relative paths which will break the links.

I'll fix the link text, so closing this

PR #5114, opened 2024-01-19
Update solutioncomponent.md

fordosa90:
Add missing component types, which have fixed value (unlike Bots, Settings, Connection References)

### DIFF
--- a/powerapps-docs/developer/data-platform/reference/entities/solutioncomponent.md
+++ b/powerapps-docs/developer/data-platform/reference/entities/solutioncomponent.md
@@ -159,6 +159,8 @@ These columns/attributes return false for both **IsValidForCreate** or **IsValid
 |162|Mobile Offline Profile Item||
 |165|Similarity Rule||
 |166|Data Source Mapping||
+|181|Virtual table data provider||
+|183|Virtual table data source||
 |201|SDKMessage||
 |202|SDKMessageFilter||
 |203|SdkMessagePair||
@@ -176,10 +178,10 @@ These columns/attributes return false for both **IsValidForCreate** or **IsValid
 |400|AI Project Type||
 |401|AI Project||
 |402|AI Configuration||
+|418|Data flow||
 |430|Entity Analytics Configuration||
 |431|Attribute Image Configuration||
 |432|Entity Image Configuration||
-
 
 
 ### <a name="BKMK_CreatedBy"></a> CreatedBy


JimDaly (commented):
@fordosa90 

Thanks for your contribution.  However, we don't accept changes to this generated reference documentation.

This reference documentation is generated by a program that uses the table schema definition data returned by Dataverse APIs. 

The reason these options are not in the doc is because they are not returned by the API. The next time these reference documents are generated, your changes would be overwritten.

I understand that the options in this optionset are no longer considered the source of truth for solution component values.

I'm aware that this page [Solution concepts > Solution components](https://learn.microsoft.com/power-platform/alm/solution-concepts-alm) says this:

> To view a list of component types that can be added to any solution, see [ComponentType Options](https://learn.microsoft.com/en-us/powerapps/developer/common-data-service/reference/entities/solutioncomponent#componenttype-options).

I think that page needs to be updated to describe how people should get these values.

fordosa90:
@JimDaly thank you, appreciate the quick and great feedback!

My concern is, if in this case this document is not the source of truth, then I am not sure what is, as exactly that was my drive to update this document to keep it relevant. 

The components that I have included are known to mee as I saw them across multiple tenants and instances in the solution files, and could not find relevant documentation for them. So I am not sure what would be then the official way on "how people should get these values" if not trough this page itself.

That said, indeed if this page is automatically generated, then my updates should not belong here either as they would be simply overwritten.

Any suggestion?

JimDaly:
@fordosa90 

Ideally, the `solutioncomponent.componenttype` column definition should provide you with this information, but it doesn't. You aren't the first person to ask about this. I'm afraid I don't know the answer.

This fetchXml query provides a distinct list of all the integer `componenttype` values in the table:

```
<fetch distinct='true'>
  <entity name='solutioncomponent'>
    <attribute name='componenttype' alias='type' distinct='true' />
    <order attribute='componenttype' descending='true' />
  </entity>
</fetch>
```

But you will see that only those defined by the `componenttype` options metadata provide formatted value labels, the rest leave you guessing. The `solutioncomponent` table doesn't have primary name column, so there is no information to tell you what that row is for.

As I mentioned above, I think this ought to be explained in this article: [Solution concepts > Solution components](https://learn.microsoft.com/power-platform/alm/solution-concepts-alm) instead of the incorrect information that is there today.  I encourage you to add a doc feedback issue on that page so that the owners of that documentation can improve it. 

But I also understand that you might not need the component type value. I have seen it used together with the [AddSolutionComponent](https://learn.microsoft.com/en-us/power-apps/developer/data-platform/webapi/reference/addsolutioncomponent?view=dataverse-latest) message, where the `ComponentType` parameter requires it. You use this message to add a component to a solution after the fact.

But there are other ways to add a component to a solution using code. I think the preferred pattern is to [Associate a solution component with a solution](https://learn.microsoft.com/en-us/power-apps/developer/data-platform/optional-parameters?tabs=sdk#associate-a-solution-component-with-a-solution) using an optional parameter or with an parameter that creates or updates the item.  With this pattern, you don't need to know the `componenttype` value.  Does this represent a possible alternative for you?

I need to close this pull request without merging your changes. 

I encourage you to create a doc feedback issue for the [Solution concepts > Solution components](https://learn.microsoft.com/power-platform/alm/solution-concepts-alm) or [ Work with solutions using the Dataverse SDK > Add and remove solution components](https://learn.microsoft.com/en-us/power-platform/alm/solution-api#add-and-remove-solution-components) so that the solutions team can hear from customers that this is causing friction. You can include a link to this closed PR.

PR #5141, opened 2024-01-29
Update set-up-timeline-control.md

ericregnier:
Added limitations section and limitation on user table


Mattp123:
@ericregnier  - User forms don't support attachments or notes, which is a primary feature of the timeline. Feel free reopen this request and provide your use case as to using the timeline with a User table form in the comments of the PR.
Closing without merging.

ericregnier:
@Mattp123, this is the reason I added the limitation to make it clear for the community since it's not supported on User forms. It would definitely be a good feature to support, and idea submitted here: https://ideas.powerapps.com/d365community/idea/d1a1e6f0-b1be-ee11-92bd-000d3ae54be7

I believe there are many use cases, model-driven apps are not always contact/account centric so having the ability to add/view activities related to a User would be nice, the same way and  reasons we enable it on other OOB and custom tables...

PR #5209, opened 2024-02-13
Update mobile-deep-links.md

trdehove:
providing an example of a model-driven app deeplink as customers tend to keep https:// in the orgurl


mduelae:
Closing public pr and reopen in private repro

PR #5214, opened 2024-02-13
Update customize-list-form.md

fselkirk:
View access to the site is not required to use the form. Shares can be at the list level. This is correct in the FAQ ("the form inherits permissions from the list"), but was overstated in Requirements.


mduelae:
closing- content has been moved/updated

PR #5242, opened 2024-02-28
Update dataverse-sql-query.md update version, preview and size limts

RichdiMSFT:
Removed the top level preview tag as the feature is generally available but only SSMS isn't fully supported.  Remove the SSMS version specific issue as it is fixed Change 80MB size limit to 5 mintue timeout.


phecke (approved):
Small edits made to changed text

PR #5278, opened 2024-03-21
Update isValid.md

ericregnier:
Added more details of how the isValid function is determined including an example

### DIFF
--- a/powerapps-docs/developer/model-driven-apps/clientapi/reference/attributes/isValid.md
+++ b/powerapps-docs/developer/model-driven-apps/clientapi/reference/attributes/isValid.md
@@ -31,7 +31,8 @@ All
 
 **Type**: Boolean. 
 
-**Description**: true if the column value is valid; false otherwise.
+**Description**: true if the column value is valid; false otherwise. <br />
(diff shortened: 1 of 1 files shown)


JimDaly:
@ericregnier 

Thanks for your contribution. I'm going to push this change through our internal repo rather than take your changes from this public repo.  I'll add you as a contributor to the article.

This is my revised content:

```
**Description**: `true` if the column value is valid; `false` otherwise.

The validity logic uses the properties of the [Dataverse column](../../../../data-platform/entity-attribute-metadata.md#column-types). Each type of column may have properties that limit valid values.

For example, a **Whole Number** column uses the [IntegerAttributeMetadata class](/dotnet/api/microsoft.xrm.sdk.metadata.integerattributemetadata) which provides the [MaxValue](/dotnet/api/microsoft.xrm.sdk.metadata.integerattributemetadata.maxvalue) and [MinValue](/dotnet/api/microsoft.xrm.sdk.metadata.integerattributemetadata.minvalue) properties. These properties are editable in the column definition in [Power Apps](https://make.powerapps.com) designer as **Maximum value** and **Minimum value** respectively.

If a Whole Number column has **Maximum value** of 0 and **Minimum value** of 100, the `isValid` function will return `false` when the current unsaved value on the form is 1000 and `true` whenever the current value falls between 0 and 100.
```

ericregnier:
Perfect, thanks @JimDaly!

PR #5285, opened 2024-03-27

shaheerahmadch:
Corrected the JSON Schema

### DIFF
--- a/powerapps-docs/developer/data-platform/query-json-columns-elastic-tables.md
+++ b/powerapps-docs/developer/data-platform/query-json-columns-elastic-tables.md
@@ -73,11 +73,11 @@ OData-Version: 4.0
 {
     "contoso_deviceid" : "device-001",
     "contoso_sensortype", "Humidity",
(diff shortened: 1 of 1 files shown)


JimDaly:
@shaheerahmadch 

Thanks for your contribution.

The changes you made look right to me. 

We have sample code at [GitHub PowerApps-Samples/dataverse/webapi/C#-NETx/ElasticTableOperations/](https://github.com/microsoft/PowerApps-Samples/tree/master/dataverse/webapi/C%23-NETx/ElasticTableOperations) which uses JObject to compose the payload, and this looks like the JObject definition got pasted in over the valid JSON that should be sent in this example HTTP request. The published example can't work because the payload isn't valid JSON.

Thanks for catching this.

JimDaly (approved):
This appears to fix errors that were introduced in this commit: https://github.com/MicrosoftDocs/powerapps-docs/commit/a1e618e18bcb56db4cbad52c164cfbdce466c531

PR #5302, opened 2024-04-18
Added note about unsupported call

fikaradz:
Added note about unsupported call


JimDaly:
Please apply this change to the internal repo

JimDaly:
Created #9442 on the internal repo

PR #5315, opened 2024-04-26
Update createRecord.md to note images in offline

eschavez:
Making a note in the createRecord to mention that images are not supported in offline mode.


JimDaly:
Responded via email and closing.

PR #5318, opened 2024-05-02
Create getFilterXml-description.md

TKlingebiel:
Adding Description


JimDaly:
Looks like this is already included in #5320

PR #5320, opened 2024-05-02
Update of the functions available for grid controls

TKlingebiel:
Added two functions that are available since at least 2019 and not documented yet.


JimDaly:
@TKlingebiel 
Thanks for your contribution. 
Not all methods that exist in the runtime are supported. We don't document the unsupported ones.
I'll need the engineering team to agree to document these methods.

@jasongre 
Please advise whether we support these getFilterXml and setFilterXml methods and want to document them.

juergenbeck:
I also vote for opening up the ability to load data in a subgrid in a supported way. Currently in the Power Apps and Dynamics CRM community, the setFilterXml method has been used to do this for several years without issue. Perhaps this setFilterXml method is good enough to be supported like it is currently? In any case, it would be very helpful if it were possible to load the subgrid on a supported option in a form with a FetchXml-based filter to give users more options for displaying relevant data in a form.

JimDaly:
@TKlingebiel & @juergenbeck 

I heard from the dev team and they made it clear that `getFilterXml` and `setFilterXml` are not supported. These methods are not documented because they aren't supported.

Please see the important note in [Client API Reference for model-driven apps](https://learn.microsoft.com/power-apps/developer/model-driven-apps/clientapi/reference):

> Also, while debugging, you may find methods and objects in the Client API object model that aren't documented. Only documented objects and methods are supported.

So rather than a documentation request, you have a feature request to support these APIs or enable similar functionality in a different way. For feature requests, I can only direct you to the [Power Apps Ideas site](https://ideas.powerapps.com/d365community/forum/f1458c72-ae29-ed11-9db2-000d3a8c451e). @jasongre may want to track this.

Closing this because we can't accept the contribution. Sorry.

PR #5321, opened 2024-05-06
Update data-retention-set.md

manasdalaimicrosoft:
Added note highlighting the change of behavior for LTR enablement via solution import.


Mattp123:
@manasdalaimicrosoft - Thank you for your contribution. However, this content has already been added to the article from another pull request. Closing without merging.

PR #5322, opened 2024-05-06
Update data-retention-set.md

manasdalaimicrosoft:
Added notes section regarding the behavior of LTR enablement via solution import


Mattp123:
@manasdalaimicrosoft - Thank you for your contribution but this content has already been added in another pull request. Closing without merging.

PR #5324, opened 2024-05-10
Remove "Everyone" update share-app.md

arjunmayur:
I've made changes to remove mention of the "Everyone" group from this document. Microsoft is recommending that makers do not share their canvas apps with "Everyone" due to the overexposure that comes with sharing with that group. We are planning to eventually phase out the ability to share with "Everyone".


arjunmayur:
Hi, 

We are planning to make changes so that sharing with 'Everyone' will be blocked by default within the next 2-3 weeks. When can we expect this to be reviewed and merged?

Thanks!

PR #5326, opened 2024-05-15
Fix wrong return type in getEventSource example

Jaronem:
The OnChange event returns formContext.data.entity.attributes and not formContext.data.entity (as also seen in the event table below).


JimDaly:
@Jaronem 

Thanks for your contribution. This looks like a clear miss.

But I think it should say:

> ..this method returns an item from the **formContext.data.entity.attributes** collection that represents the changed column.

I've made that change in a separate commit, so closing this one.

PR #5327, opened 2024-05-20
Update performance-insights-overview.md

komala2019:
We are no more supporting the performance insights feature for production.


Mattp123:
This article has been removed because the feature is no longer available. Closing without merging.

PR #5328, opened 2024-05-20
Update create-edit-web-resources.md

komala2019:
We have an IcM whether the customer has created a custom UI using web resources which is a viable solution path for the customer but the design architecture of the web resources doesn't guarantee every scenario to work as expected as the design was never meant to support custom UI.


Mattp123:
Submitted by internal team member. Questions to clarify this PR went unanswered. Original submitter has left the team and backfill has not responded.
Closing without merging.

PR #5338, opened 2024-06-18
Update platform library documentation to fix v9 npm name and attempt to explain ranges

jopursle:
Earlier today, I noticed that the npm name for Fluent v9 was pointing to the v8 import, so updating the docs to use the `@fluentui/react-components` name for v9.
Also, based on email discussion, attempting to explain what version of React is loaded when version is specified and provide the fluent v9 supported ranges.


JimDaly:
Moved changes to https://github.com/MicrosoftDocs/powerapps-docs-pr/pull/10231

PR #5339, opened 2024-06-21
Changed a single character from uppercase to lowercase to get code working.

JamesNJep:
When setting the TextInput Mode property to multiple lines. It appears to only work when the L is lower case. Changed "MultiLine" to "Multiline" in documentation.


mduelae:
fixed in internal PR

PR #5340, opened 2024-06-21
Fixed grammar issue.

Stoefza:
Fixed a grammar issue


maustinjones:
This particular change doesn’t read properly but the statement could be rewritten in a clearer and more concise manner. So I’m supportive of improving this.

mduelae:
fixed in internal PR

PR #5342, opened 2024-07-02
Clarification of flows association with apps

Ravi-Chada:
During export and import of a canvas app, a flow is not part of that process. This clarification is added based on the feedback from Nokia


Ravi-Chada:
Closing this to make updates

PR #5343, opened 2024-07-04

shaheerahmadch:
Added:

If you want to retrieve grouped rows based on a specific field without aggregating them, you need to select distinct rows. However, Cosmos DB SQL API does not support the traditional DISTINCT keyword in the same way as SQL Server or other RDBMS.

### DIFF
--- a/powerapps-docs/developer/data-platform/query-json-columns-elastic-tables.md
+++ b/powerapps-docs/developer/data-platform/query-json-columns-elastic-tables.md
@@ -209,7 +209,8 @@ OData-Version: 4.0
 ```
 
 
-If you want to retrieve grouped rows based on a specific field without aggregating them, you need to select distinct rows. However, Cosmos DB SQL API does not support the traditional DISTINCT keyword in the same way as SQL Server or other RDBMS.
+Note:
+Cosmos DB SQL API does not support the traditional DISTINCT keyword in the same way as SQL Server or other RDBMS.
 
 ---
 
--- a/powerapps-docs/developer/data-platform/query-json-columns-elastic-tables.md
+++ b/powerapps-docs/developer/data-platform/query-json-columns-elastic-tables.md
@@ -208,6 +208,9 @@ OData-Version: 4.0
 }
 ```
 
+
+If you want to retrieve grouped rows based on a specific field without aggregating them, you need to select distinct rows. However, Cosmos DB SQL API does not support the traditional DISTINCT keyword in the same way as SQL Server or other RDBMS.
+
 ---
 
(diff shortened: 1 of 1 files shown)


JimDaly:
@sumantb-msft 

Can you review the changes that @shaheerahmadch added to this document?

This [Query JSON columns in elastic tables article](https://learn.microsoft.com/power-apps/developer/data-platform/query-json-columns-elastic-tables?view=dataverse-latest&tabs=sdk) is specifically about using `ExecuteCosmosSQLQuery`.

The new content added to the bottom of the article is:

> If you want to retrieve grouped rows based on a specific field without aggregating them, you need to select distinct rows. However, Cosmos DB SQL API does not support the traditional DISTINCT keyword in the same way as SQL Server or other RDBMS.

I'm not clear about how this content helps the reader. It essentially says:

> If you want to <Do X>, you need to <Do Y>. However, Cosmos DB SQL API does not support <Doing Y>.

What value does this offer to the reader other than saying you can't _<Do X>_?
If there is no solution, just let people know what they can't do so that they won't waste time trying to do it?

sumantb-msft:
I am not sure what is the use case this newly added content is clarifying. I will need more details about the use case this is talking about before approving this content.

powerapps-docs/developer/data-platform/query-json-columns-elastic-tables.md
```diff
@@ -208,6 +208,9 @@ OData-Version: 4.0
 }
 ```
 
+
```
JimDaly, line 211: Please provide more details about the use cases for this content.

JimDaly (changes requested):

shaheerahmadch:
Updated

JimDaly:
@shaheerahmadch 

Thanks for your update. I see the comment now simply says:

> Note:
> Cosmos DB SQL API does not support the traditional DISTINCT keyword in the same way as SQL Server or other RDBMS.

We don't want to comment on the capabilities of Cosmos DB SQL API in the Dataverse documentation. Our feature uses that API, but we don't own it. We shouldn't make any statements about what that API does or doesn't support.  We have a link to [Learn about queries in Azure Cosmos DB for NoSQL](https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/query/getting-started), and that is all we plan to say about the Cosmos DB SQL API.

If our feature doesn't support some Cosmos DB SQL API capability that people might expect, that is something I think we are responsible to explain. But that doesn't appear to be the case here.

Make sense?

I'm going to close this. Please feel free to respond if I've fully misunderstood the point.

PR #5344, opened 2024-07-04

shaheerahmadch:
## Added Information about pagging:

To implement pagination in a Cosmos DB SQL query, you typically use a combination of the OFFSET and LIMIT clauses (similar to SQL Server’s OFFSET FETCH). However, Cosmos DB SQL API has its own way to handle pagination via continuation tokens rather than OFFSET and LIMIT directly.

Here is how you can approach it:

First Request: Fetch the initial set of records along with a continuation token.
Subsequent Requests: Use the continuation token to fetch the next set of records.
Here is an example of how you can perform this:

continue....


shaheerahmadch:

PR #5350, opened 2024-07-15
Update file-column-data.md

daryllabar:
In my testing the file id and the file_name columns were both included when utilizing the AllColumns flag for the Query Expression.  As this is the default behavior of the AllColumns flag, IMHO this not should be removed.


JimDaly:
@daryllabar 
Thanks for your contribution.

@trin-msft 
I tested this and the error no longer occurs. Something must have changed since this was written. See my email.

Please approve this PR and I'll merge it.

trin-msft (approved):

PR #5358, opened 2024-08-09
fix: c# code sample causes duplicate first page

tdashworth:
The `page` variable only gets updates _after_ its used in the while loop causes the page numbers it go: 1, 1, 2, 3...


JimDaly:
@tdashworth 
Thanks for notice this and submitting this PR.

I've verified that the change applied works and updated the sample in another branch.
I also tested the QueryExpression version of the RetreiveAll sample and added the ability to specify page size.

Since the changes are in another PR I will deploy today, closing this one.

PR #5360, opened 2024-08-13
Update wrap-how-to.md

code2exe:
This change is to fix the issues faced when authenticating to MS Graph without proper authorization scopes. This can be seen in the screenshot below.

image

When the proper authorization scopes are added, here is the result below:

image


code2exe:
@tapanm-MSFT This PR is still pending your attention.

Thank you

PR #5362, opened 2024-08-20
Sample App Update

DigitalKrony:
Remove mention of model-driven-app Samples within the documentation.


DigitalKrony:
Per a requirement, based on issues with Signed dll files and outdated Sample apps, these connections and the corresponding documentations need to be removed. 

There is a sibling PR out within PPUX to feature flag and disable the display of these Sample Apps until such time they can be created in a manner that is beneficial to our user base.

jamesdeancalhoun (approved):
Needed since removing App Module sample apps

Mattp123:
@DigitalKrony - Thank you for your contribution. The article has already been removed via another pull request. 
Closing without merging.

PR #5417, opened 2025-02-12
Correction to the format of default config, specifically the toolbar setting

socialcpf:
The existing format of the toolbar config is incorrect.  It currently shows as an array or arrays and it should be a json object with an items attribute that is an array of the toolbars.


socialcpf:
@microsoft-github-policy-service agree [company="Strathcona Solutions"]

PR #5433, opened 2025-04-11
Update aggregate-data.md

daryllabar:
The Sample data is wrong for the date grouping so that needed to be fixed. There was no information on how to access the returned values, which was very confusing to me.  Added details that they are returned as an Aliased value, and that the column will not be returned at all if it is a null grouping value.


JimDaly:
Thanks Daryl 

I'm including your changes in another PR. I see the value that you added but I want the information in different places. Should be published by EOD.

But that means I'm closing this PR.
I added you to the list of contributors in the metadata.

PR #5434, opened 2025-04-12
Update generate-early-bound-classes.md

daryllabar:
I get a lot of feedback from developers who don't use the EBG because they want to generate the code via a command line a build pipeline.  I think it's important to call out that capability here, as well as it providing additional features that the pac modelbuilder doesn't. (transliteration is especially important for non English  character sets)


JimDaly:
Thanks @daryllabar .

I've edited your changes in another branch and merged them.

PR #5436, opened 2025-04-16
Update data-platform-create-business-rule.md

powerappsviking:
Added information about how business rules do not work with all actions on bulk-edit and data import jobs. If this should be moved to https://learn.microsoft.com/en-us/power-apps/developer/model-driven-apps/clientapi/events-forms-grids?tabs=add-event-handlers-unified-interface#bulk-edit-forms instead then I can update it there. But the table information seemed like a more appropriate place for it.


powerappsviking:
> @microsoft-github-policy-service agree [company="{your company}"]

N-Kannan (approved):
Okay, this looks good.

powerappsviking:
@Mattp123 @tapanm-MSFT let me know if you need me to add/change anything on this

PR #5450, opened 2025-06-10
Update sql-server-security.md

jelopezf:
Update to include more auth types and make clear that service principal connections added to apps prior to May 2025 should be recreated


EllenWehrle (dismissed):

EllenWehrle:
@mduelae, please review and approve if it's ready to publish.

PR #5454, opened 2025-06-19
Update multitable-lookup.md

ThatGuyGW:
Add note to clarify behaviour of displayed tables to the user.

### DIFF
--- a/powerapps-docs/developer/data-platform/webapi/multitable-lookup.md
+++ b/powerapps-docs/developer/data-platform/webapi/multitable-lookup.md
@@ -28,9 +28,6 @@ power to define any other multi-table lookups they might need.
 > At this time users can create and modify custom multi-table lookups via the SDK or Web APIs.
 > Interactive user interface support will be coming in a future release.
 
-> [!NOTE]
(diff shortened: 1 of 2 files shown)


JimDaly (changes requested):
See my other comments about this PR

ThatGuyGW:

ThatGuyGW:
@JimDaly That's a fair point, I've made the requested changes now. Just RE that page though, I'm unsure if that's the right place either, since the Multi-Table lookup doesn't seem to be it's own control, it just uses the standard lookup control. Granted it has the additional UI bit for selecting different tables but the Customer lookup has that as well and that's not in that section either.

Also I'm unsure if Option Set should be in that list either as it's just describing the default control, not additional, unless I'm misunderstanding the point of that page.

Either way let me know if you're happy with the added section, I tried to keep it similar length to other entries but can add more detail if required.

JimDaly:
We updated these sections of the docs to include missing information about multi-table (polymorphic) lookups. 

- [Different types of lookups](https://learn.microsoft.com/en-us/power-apps/maker/data-platform/types-of-fields#different-types-of-lookups)
- [Multi-table lookups](https://learn.microsoft.com/en-us/power-apps/maker/model-driven-apps/form-designer-add-configure-lookup#multi-table-lookups)


These changes should address the root cause for the proposed changes in this PR, so closing this PR.

PR #5461, opened 2025-07-10
Update add-app-solution-default.md

mattruma:
Remove preview settings, it is no longer marked as preview with in the Power Platform > Environment > Settings > Features any longer.


powerapps-docs/maker/canvas-apps/add-app-solution-default.md
ChrisGarty, line 114: This line should have the preview mention removed as well

ChrisGarty (changes requested):

PR #5474, opened 2025-09-26
Change DefaultValue from False to True

crisp555:
Updated for [Incident 664429282](https://portal.microsofticm.com/imp/v5/incidents/details/664429282/summary) : The default of "Add cloud flows to a solution by default" value is different with Microsoft Learn site


crisp555:
This seems to be an auto-generated page.  Could you help me find the owner who can re-generate this page?

PR #5487, opened 2025-11-21
Revise download instructions and file handling

AliAAdamu:
Updated instructions for downloading scripts and handling files for Azure Synapse Link.
The current link points to nowhere :)


Mattp123:
The article and procedures have changed so these updates no longer apply. Closing without merging.

PR #5494, opened 2026-01-25
Update ForAll + Patch examples

watana2:
Updated ForAll( ) + Patch ( ) examples to improve clarity on performance.


samt-source (approved):
Strong clarification. Explicitly highlighting the sequential request behavior of Patch() inside ForAll() improves technical accuracy and is especially important for enterprise and government-scale implementations where performance and service reliability are critical. This update strengthens the guidance for makers working in high-volume environments. Approving.

PR #5501, opened 2026-02-20
Update create-edit-powerbi-embedded-page.md

p-ciprian:
Misleading Information about the PowerBi Control for dataverse Model Driven App forms.

Misleading content from MS on how to integrate PowerBI with OOB form components. 

However, this component was in preview, and withdrawn, but Microsoft removed any trace of this. 

After few hours of investigation i identified this text https://github.com/MicrosoftDocs/powerapps-docs/commit/be92f82a69fb0b139bc226dc6f46de309f914981.  

MS Video still on youtube: https://www.youtube.com/watch?v=ZuXAqlQdOb4 

Third-party sample[: How to add Power BI Tiles to your D365 Form | Power Apps Tutorial](https://www.youtube.com/watch?v=f8dXqLMnNto)  

Resolution: Update documentation, clearly state this feature does not exist

### DIFF
--- a/powerapps-docs/maker/model-driven-apps/create-edit-powerbi-embedded-page.md
+++ b/powerapps-docs/maker/model-driven-apps/create-edit-powerbi-embedded-page.md
@@ -24,6 +24,9 @@ Read more about relevant Power BI licenses at [Power BI pricing](https://powerbi
 
 To learn more about Power BI reports and dashboards, go to [Create reports and dashboards in Power BI](/power-bi/create-reports/).
 
+> [!IMPORTANT]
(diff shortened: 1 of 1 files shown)


JimDaly:
@p-ciprian 

Thanks for your contribution, but we can't accept it 'as-is' and I'll close this pull request.

I think I understand your feedback is more than just this change.

With some investigation I found that there was a Power BI control released as a preview feature. The feature was later withdrawn and was never generally available.

When we deprecate a feature, we publish information in a page like [Important changes (deprecations) coming in Power Platform](https://learn.microsoft.com/power-platform/important-changes-coming). This provides some way for people to discover that the feature existed.

The Power BI control preview feature was never fully released, so we can't deprecate it. The page about it was removed and the last mention of it in our docs went with it. As you mentioned - it is still mentioned in videos on YouTube.

Agree with your recommendation to preserve some way for people to discover preview features that never made it to GA so that people can discover this fact, but adding the change in this PR isn't the way we need to start doing this.

PR #5502, opened 2026-03-02
Enhance ILogger documentation with telemetry key setup

gfugy:
Added instructions for updating the telemetry instrumentation key in Dataverse environment.


gfugy (commented):
Corrected markdown syntax

JimDaly:
@gfugy In the email discussion we concluded that the steps described in this PR are not necessary. Instead people should only use the instructions found in [Create an export package](https://learn.microsoft.com/power-platform/admin/set-up-export-application-insights#create-an-export-package). This is the only correct way to set this. 

As documented, the person setting this must have Administrator, Contributor, or Owner access to the Azure Application Insights resource.

PR #5511, opened 2026-04-22
Fix command option from flowId to flow-id

dudelis:
The command gives an error when running it with -flowId paramter, so i changed that to flow-id.


eschavez (approved):

JimDaly (approved):

JimDaly:
@dudelis Thanks for your contribution.

PR #5515, opened 2026-04-26
Correct relationship direction in navigation properties

Mosh-K:
Both intro sentences for the two navigation property types had the relationship direction inverted, contradicting the prose immediately below them.

- **Line 39** (single-valued): A single-valued navigation property points from many records to one referenced record. Line 53 right below describes this correctly: "connects multiple `account` records to a single `systemuser` record."
- **Line 116** (collection-valued): A collection-valued navigation property points from one record to a collection of related records. Line 128 right below explicitly says "regardless of whether the relationship is one-to-many or many-to-many."

Update: also corrected the H3 subsection at line 130 ("Many-to-one relationships" → "One-to-many relationships") and its opening sentence. The subsection lives under "Collection-valued navigation properties" (the one-to-many side), so the heading was inverted in the same way.

### DIFF
--- a/powerapps-docs/developer/data-platform/webapi/web-api-navigation-properties.md
+++ b/powerapps-docs/developer/data-platform/webapi/web-api-navigation-properties.md
@@ -36,7 +36,7 @@ The following table describes the attributes of `NavigationProperty` elements.
 
 ## Single-valued navigation properties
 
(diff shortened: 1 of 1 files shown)


JimDaly:
@Mosh-K 
Thanks for your contribution. This has been incorrect for a long time. Thanks for changing it.
You will probably find more like this.

PR #5517, opened 2026-05-14
Fix `Use a collection` section in working-with-variables instructions

mattbaker-digital:
Corrected broken syntax and add missing step for adding a label. Also corrected numbering sequence.


mattbaker-digital:

PR #5520, opened 2026-06-06
Add SQL example for sorting with JOIN

rappen:
Added SQL example for sorting by joined columns.


JimDaly (approved):

JimDaly:
@rappen Thank you for your contribution.

PR #5521, opened 2026-06-09
Add special characters limitation to Function creation guidance

allandecastro:
## Summary

Adds a **Special characters limitation** subsection to the *When to create a Function* section in the Custom API article.

Custom API Functions use an HTTP `GET` request that passes all parameter values in the URL. As a result, string parameter values containing certain OData-reserved characters (`/ < > * % & : \ ? +`) aren't supported and cause `400 Bad Request` or `404 Not found` errors. This is the same OData parser limitation already documented for alternate keys.

The new subsection documents the unsupported characters, the resulting errors, links to the existing alternate-key limitation for context, and recommends creating an *Action* (POST with body parameters) when these characters must be supported — noting that the Function/Action choice can't be changed after creation.

## Change
- `powerapps-docs/developer/data-platform/custom-api.md` — new `### Special characters limitation` subsection under *When to create a Function*.

Rendered section: https://learn.microsoft.com/en-us/power-apps/developer/data-platform/custom-api#when-to-create-a-function


JimDaly:
Allan -

Thanks for taking the time to create this Pull request. Your question and this PR gave the dev team something to review and consider. They tested it and found a workaround

It was already documented, but clearly not discoverable. I've started some changes to make this easier to find.

In [Passing parameters to a function](https://learn.microsoft.com/power-apps/developer/data-platform/webapi/use-web-api-functions#passing-parameters-to-a-function) we have content that describes the workaround. If the special characters are in the URL path segment, OData will reject them. But they should work if passed as query string parameters.

The content we have had for years never connected the dots with the very similar constraint that alternate keys have. The team tested this and found this workaround won't work for alternate key values that contain the special characters.

I've prepared another PR on our internal repo that addresses both this workaround and another workaround for special characters in alternate key values (basically you can use `$filter` to retrieve the single value)

Thank you for creating this PR, but I'm going to close it in favor of another one that will get pushed through from our side soon.

allandecastro:
Hi @JimDaly ,

No problem closing it — I appreciate you taking it in a different direction.
One thing though: you mentioned the workaround is passing parameters as query string parameters, but in practice we tested that approach and custom API functions still failed. Here's what we tried:

`GET /api/data/v9.2/CustomAPIFunction(param1=@p1,param2=@p2)?@p1=''&@p2='value&with&ampersand'`

Even with the parameter alias syntax (following the documented best practice), the API returned 400/404 errors. We also tried URL-encoding the ampersand in the query string (@p2='value%26with%26ampersand'), but that failed too.

So while the workaround might work for standard Dataverse operations, the constraint appears to be broader for custom API functions specifically.

JimDaly:
@allandecastro 

The following is a snippet from a report that the developer generated while testing this:

---

  <h3>Confirmed on a real Custom API function</h3>
  <p>To verify this isn&rsquo;t specific to the OOB function, a real Custom API <strong>function</strong>  <code>new_FindThing</code> was created &mdash; unbound, active ( <code>statecode 0 / statuscode 1</code>), with a string parameter  <code>Name</code> and a string response property  <code>Result</code> (a function must declare a response to be a valid OData function). Once configured, it is callable. This is the MVP&rsquo;s exact scenario ( <code>msc_findCustomers(name='&hellip;')</code>):</p>
  <table>
    <thead>
      <tr><th>Call</th><th>Result</th><th>Meaning</th></tr>
    </thead>
    <tbody>
      <tr>
        <td> <code>new_FindThing(Name='PLAIN')</code> &mdash; benign, inline</td>
        <td><span class="pill ok">200 OK</span>  <code>{"Result":null}</code></td>
        <td>Parses, routes, executes (no plugin attached, so  <code>Result</code> is null &mdash; irrelevant here)</td>
      </tr>
      <tr>
        <td> <code>new_FindThing(Name='M%26M')</code> &mdash; special, inline</td>
        <td><span class="pill bad">400 BadRequest</span></td>
        <td>The MVP&rsquo;s failing case &mdash; dies at the <strong>IIS / URL&#8209;path parser</strong>, before OData routing</td>
      </tr>
      <tr>
        <td> <code>new_FindThing(Name=@p)?@p='M%26M'</code> &mdash; <code>@param</code></td>
        <td><span class="pill ok">200 OK</span>  <code>{"Result":null}</code></td>
        <td><strong>The fix</strong> &mdash; the special value  <code>M&amp;M</code> is delivered to the real Custom API and it executes</td>
      </tr>
    </tbody>
  </table>
  <div class="callout ok">
    <strong>The same fix works on an actual Custom API</strong>
    Inline special char &rarr; <span class="pill bad">400</span> (parser layer); the identical call via <code>@param</code> &rarr; <span class="pill ok">200</span> with the value delivered. This confirms the behavior is not specific to OOB functions &mdash; it applies to <em>every</em> Custom API function with a string parameter, exactly as the MVP reported, and parameter aliases resolve it.
  </div>
  
  ----
  
  Do you see any difference between what he tested and what you are reporting?
  
  I will also ask him to double-check his work. We should both get the same results.

allandecastro:
Thanks @JimDaly — that test report is really helpful, and it matches what we eventually landed on.

I've pushed an update to this branch that reworks the **Special characters limitation** section to document the actual workaround rather than just steering people to Actions. The key points:

- The inline form fails as expected: `sample_FindThing(Name='M&M')` → `400 Bad Request`.
- The reliable workaround is **parameter aliases with URL-encoded values** — the value is decoded before it reaches the plug-in:
  ```http
  GET [Organization URI]/api/data/v9.2/sample_FindThing(searchKey=@p1,name=@p2)?@p1=''&@p2='M%26M'
  ```
  The plug-in receives `name = "M&M"`.
- Added an encoding reference (`&`→`%26`, `/`→`%2F`, `+`→`%2B`) and kept the *use an Action instead* path as the alternative.

Happy to adjust naming/format to fit the house style. Let me know if you'd like any changes.

JimDaly:
@allandecastro 
We pushed out a few updates yesterday via our internal paired repo. See [Update due to special characters in functions (#12137)](https://github.com/MicrosoftDocs/powerapps-docs/commit/356071d812369db39fb62176f5a88cc12e836de3#diff-5e2c3e00abfdfaa6f70ef75e4f3ede86a9ac4201c41f13a3930419b7397c320b)  so, there would be some redundant information with the current changes.

The thing this change doesn't include is to encourage people to consider using actions rather than functions. If you can integrate this into the updated content, then I could push this through.

allandecastro:
Thanks Jim — that clears up the visibility gap. I can see your commits land in the public repo but not the internal PR discussions, so I hadn't realized #12137 already published the parameter-alias workaround note. Makes sense now.

I've reworked this branch to fit on top of that. I merged the latest `main`, removed my "Special characters limitation" subsection entirely (it now duplicates the note you shipped in #12137), and added a single bullet right alongside it with the one piece that wasn't covered:

> To avoid the special-character handling that functions require, consider creating your custom API as an Action instead. An Action uses an HTTP `POST` request that passes parameter values in the request body, so string values that contain special characters need no parameter aliases or URL encoding. Because you can't change whether a custom API is a Function or an Action after you create it, decide which one you need early in your design.

The net diff against `main` is now just that one line — no overlap with #12137. Happy to adjust the wording to match house style if you'd prefer. Thanks again for sticking with this to get it landed.

JimDaly (approved):
@allandecastro Thanks for your contribution!

PR #5526, opened 2026-08-10
Document generally available Xrm.Copilot methods

MitiJ:
## Summary

- Separate generally available and preview Xrm.Copilot methods into distinct tables
- Mark `updateContext` as generally available using standard documentation conventions
- Keep `executeEvent` and `executePrompt` labeled as preview

## Validation

- Confirmed the preview titles, notices, and descriptions remain in place for `executeEvent` and `executePrompt`
- Confirmed no preview labeling remains on `updateContext`


JimDaly:
Changes from Microsoft should pass through a different process. @MitiJ I'll follow up with you via email.

PR #5531, opened 2026-09-10
Fix variable assignment and update GH_TOKEN usage

markjulmar:
Because `CASCADE_PAT` is present in that step’s environment, injected commands can read and exfiltrate it despite GitHub’s log masking. Anyone with permission to invoke `workflow_dispatch` could exploit this.

### DIFF
--- a/.github/workflows/cascade-include-changes.yml
+++ b/.github/workflows/cascade-include-changes.yml
@@ -40,7 +40,7 @@ jobs:
         id: changed
         run: |
           if [ "${{ github.event_name }}" = "workflow_dispatch" ]; then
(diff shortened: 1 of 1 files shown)


markjulmar:
@nathlaroche This is to fix an IcM we got against this workflow.

workflow_dispatch input inputs.changed_files is interpolated directly into a run-block (FILES="${{ inputs.changed_files }}", line 43), letting an authenticated write-access collaborator inject shell commands that execute with the repo-scoped CASCADE_PAT (line 53) in scope — arbitrary code execution on the runner plus cross-repo PAT theft. True positive, Important.

nathlaroche (approved):

PR #5536, opened 2026-09-15
Fix contradictory statement about out-of-box DateOnly column behavior

Mosh-K:
## Summary

- The "Change the behavior of a date and time column" section listed the same 11 out-of-box columns (`anniversary`, `birthdate`, `duedate`, `estimatedclosedate`, `actualclosedate`, `finaldecisiondate`, `validfromdate`, `validtodate`, `closedon`, `expireson`) as both:
  1. Locked to `DateOnly` (`CanChangeDateTimeBehavior = False`), **and**
  2. `UserLocal` with `CanChangeDateTimeBehavior = True`, changeable to `DateOnly`

  These are mutually exclusive claims about the same set of columns.

- Traced the origin via git history: this content was migrated wholesale from the legacy Dynamics 365 Customer Engagement (on-premises) docs in 2018 (`d72639eeab`). The original sentence was conditional — *"However, if these out-of-box date and time attributes belong to an organization that is upgraded to CRM Online 2016 Update or CRM 2016 (on-premises), the behavior of these attributes is set to `UserLocal`..."* — but the migration dropped that qualifying clause, leaving an unconditional sentence that directly contradicts the one describing the default state above it. Every later edit (2021 terminology pass, 2023 revamp, 2026 copyedit) only reworded the sentence and never caught the contradiction.

- Verified against this repo's generated entity reference docs (`reference/entities/contact.md`) and the `dynamics-365-customer-engagement` repo (`ce/developer/reference/entities/*.md`, which cover Invoice/Lead/Opportunity/Product/Quote): all 11 columns report `DateTimeBehavior: DateOnly` and `CanChangeDateTimeBehavior: False`. None match the removed sentence.

## Change

Removes the stale, unconditional sentence:

> The behavior of these columns is set to `UserLocal` and the `DateTimeAttributeMetadata.CanChangeDateTimeBehavior` managed property to `True`, and you can change the behavior of these columns to `DateOnly` only. No other behavior transitions are allowed.

It described an on-premises CRM 2015/2016-upgrade scenario with no equivalent in modern (cloud-only) Dataverse, and contradicts verified metadata for every column it claims to describe.

## Test plan

- [ ] Docs build/lint passes
- [ ] Manual read-through for section flow after removal

🤖 Generated with [Claude Code](https://claude.com/claude-code)


JimDaly:
@Mosh-K 
Great catch. Thanks for helping to make the documentation better.

